PR #37, opened 2018-07-30
Move public env vars to travis file

kylemh:


kylemh:
Didn't do what I thought it would 😬

PR #43, opened 2018-07-31
Deploy static bundles on PRs

kylemh:
# Description of changes
Instead of building full-blown deployments for pull requests, where we're limited to 3 concurrent instances that can sleep, we're going to deploy static builds. They don't have the performance boost that Next's SSR provides, but they're just previews. In exchange, we can have unlimited static previews and they don't sleep.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ after_script: greenkeeper-lockfile-upload
 script:
   - yarn lint
   - yarn test
-  - yarn build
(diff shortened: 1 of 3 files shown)


kylemh:
Deployed preview for updated Storybook instance

Built with commit 395c3cf975f069c093f918486b99f4f7075cdf48

https://app.netlify.com/sites/operation-code-storybook/deploys/5b601572c6aed66d045e02ff

PR #44, opened 2018-07-31
Use static builds for PR Now integration

kylemh:
# Description of changes
Instead of building full-blown deployments for pull requests, where we're limited to 3 concurrent instances that can sleep, we're going to deploy static builds. They don't have the performance boost that Next's SSR provides, but they're just previews. In exchange, we can have unlimited static previews and they don't sleep.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_deploy: yarn add now # Install Now CLI
 
 deploy:
   - provider: script # Run a custom deployment script which we will define below
-    script: yarn export && now --public --static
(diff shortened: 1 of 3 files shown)


kylemh:
**This is a bot**<br />🎉 Deployed Storybook preview! 🎉<br />Click the link at the bottom of this comment, and then click **"Preview deploy ->"** to see it live.

Built with commit b3f5deccac2cddbfea301bd4947b52b05fc914fb

https://deploy-preview-44--operation-code-storybook.netlify.com

Issue #170, opened 2018-10-01
Reimplement /code_schools page


kylemh:
# Feature
## Why is this feature being added?

https://operationcode.org/code_schools exists and it needs to exist in the new code base as well.

## What should your feature do?
- Implement the code_schools page via `pages/code_schools.js`
- Emulate [this mock](https://projects.invisionapp.com/share/KSO12EFRU7E#/screens/323736695_OC_Coding_Schools_Landing_Page).
- [See live source code.](https://github.com/OperationCode/operationcode_frontend/tree/master/src/scenes/home/codeSchools) Feel free to copy-paste a lot.
- Refactor as required to match our new standards (linting, using components where possible, etc.)
- Implement new card design
- Reference moocs from API ([see here](https://github.com/OperationCode/operationcode_backend/pull/250))

kylemh:
This is a request for a new or redesigned page.

- The page should be implemented by creating a file (or editing the file if it already exists) in the `pages` folder. The file name should match the route (ex: code_schools --> `pages/code_schools.js`).
- Please remember to leverage the `<OutboundLink>` component when dealing with external links (leaving our hostname - operationcode.org)
- Do your best to limit stylesheets imported into `pages/styles`. 
- Only modularize potentially reusable content via `components` - it's okay to implement everything in the page file and let us as staff members sort out where you could write for reuse (if anywhere).
- If no design is missing mobile-specific or desktop-specific styling, do your best to work with what's given. A good rule of thumb is that any lists displayed in rows on desktop mocks, can be displayed one-per-row on mobile.

kylemh:
Please note: https://github.com/OperationCode/front-end/issues/225

markchernov:
I'd like to start working on this story. It looks like https://operationcodeapi.docs.apiary.io/#reference/0/codeschool-collection/list-all-codeschool-members?console=1   returns invalid json response. Validation result could be seen at https://jsonlint.com/

markchernov:
Is there a way to add additional schools of different types to that mock endpoint?

kylemh:
@markchernov are you on Slack?

We can chat about this issue in more detail there. I wouldn't be setting you up for success in letting you work on this issue just yet, as I don't have good enough reqs laid out.

We can chat through them and relay that info here, but this issue is a large amount of work.

markchernov:
Page Design at: 
[...]

PR #191, opened 2018-10-02
Add more tests  - 66.06%

marciojc:
# Description of changes
Add more tests #170 

100% coverage:

- Modal

- YouTubeVideo

- OutboundLink

Fix:

- Section

### DIFF
--- a/components/_common_/Modal/Modal.js
+++ b/components/_common_/Modal/Modal.js
@@ -35,7 +35,7 @@ function Modal({
   shouldCloseOnOverlayClick,
 }) {
   if (process.env.NODE_ENV === 'production' && process.env.GOOGLE_ANALYTICS_KEY) {
-    ReactGA.modalView(screenReaderLabel);
+    ReactGA.modalview(screenReaderLabel);
   }
 
   return (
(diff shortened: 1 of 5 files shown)


kylemh:

Built with commit 716d07646ab60eab9e3d8dd4cd964a8b5dd60a0d

https://deploy-preview-191--operation-code-storybook.netlify.com

components/_common_/Modal/Modal.js
```diff
@@ -35,7 +35,7 @@ function Modal({
   shouldCloseOnOverlayClick,
 }) {
   if (process.env.NODE_ENV === 'production' && process.env.GOOGLE_ANALYTICS_KEY) {
-    ReactGA.modalView(screenReaderLabel);
+    ReactGA.modalview(screenReaderLabel);
```
kylemh: oh good catch!

components/_common_/Section/Section.js
```diff
@@ -47,9 +47,8 @@ function Section({
     // Dark background colors have light heading color
     case 'gray':
     case 'slate':
-      headingTheme = 'white';
-      break;
     default:
+      headingTheme = 'white';
```
kylemh: I'd prefer this change get reverted so that we're explicit in the casings. If another theme gets added this could cause an obfuscated headache. Currently a new theme will break the component which is ideal so a dev is forced to explicitly handle the case.

marciojc: ok, no problem

kylemh:
Killing it! Thanks again!

PR #203, opened 2018-10-05
Create page script

theboynextdoor:
# Description of changes

Automate's the creation of a page inside the `/pages` directory. 

# Issue Resolved

Fixes issue requested by Kyle


kylemh:

Built with commit 455a0ec5917566a8ee9549cd110c938c75d0d6ee

https://deploy-preview-203--operation-code-storybook.netlify.com

scripts/createPage/builder.js
```diff
@@ -0,0 +1,17 @@
+const upperCaseFirstLetter = string => string.charAt(0).toUpperCase() + string.slice(1);
```
kylemh: Instead of redefining this function, go ahead and import it from `common/utils/string-utils`

scripts/createPage/builder.js
```diff
@@ -0,0 +1,17 @@
...
+
+const builder = {
+  buildJS: pageTitle => `import Head from 'components/head';
+
+export default () => {
+  const pageTitle = '${upperCaseFirstLetter(pageTitle)}';
```
kylemh: It's good that you thought to upper case it, but what happens when an individual puts a `-` or a `_` into the page name? 🤔 

- Let's throw an error if `pageTitle` includes a `-` or any empty spaces.
- Let's split the string by `_` to display in the page title variable.
- Let's put the functions which do any string manipulation into `common/utils/string-utils`.

You'll need the same amount of parsing inside the `createPage` function so the file itself isn't created.

This is a great start!

scripts/createPage/createPage.js
```diff
@@ -0,0 +1,74 @@
...
+  const filePrefix = '.js';
+
+  if (fs.existsSync(newPath + filePrefix)) {
+    console.log(`Page "${pageName}" Already Exists`);
+    return true;
+  }
+  return false;
+};
+
+const createPageTitle = pageName => {
```
kylemh: Just wanted to show a different implementation of this:
```
const createPageTitle = pageName => {
  // Example 1: `some_page_name` becomes `Some Page Name`
  // Example 2: `page` becomes 'Page'
  return pageName.split('_').map(word => capitalizeFirstLetter(word)).join(' ');
}
```

I wasn't sure what the conditional check on line 37 was for.

scripts/createPage/createPage.js
```diff
@@ -0,0 +1,74 @@
...
+  if (capitalizePageName[0] === '') {
+    capitalizePageName = capitalizePageName.slice(1);
+  }
+
+  return capitalizePageName.join(' ');
+};
+
+const createPage = (root, pageName) => {
+  // test to make sure file meets requirements
+  if (containsDashCharacter(pageName) || containsWhiteSpace(pageName)) {
```
kylemh: so going off what I said earlier on the new utility functions, here you can just do:
```javascript
const isNotValidPageName = pageName.includes('-') || pageName.includes(' ');

if (isNotValidPageName) {
  console.log(`Page: ${pageName} wasn't created because of a dash or white space in name.`);
  return false;
}

//etc
```

scripts/createPage/createPage.js
theboynextdoor: The conditional check was for if a user named their page with an underscore in the front, e.g. _foo. When splitting a string by the underscore, it will eliminate it, but when rejoining the string you will get a space in the character in front of foo, e.g. pageTitle = ' foo'.

scripts/createPage/createPage.js
theboynextdoor: Good idea. Will make that change now.

scripts/createPage/createPage.js
theboynextdoor: I will reimplement this to match your version. (A lot cleaner)

theboynextdoor: The conditional check was for this scenario: 
`// Example: "_some_page" becomes " Some Page"`

There will be an extra space before "Some"

theboynextdoor: I originally had the  `createPageTitle` function like the one above. Then I thought about this possible edge case: what if a user decides to name their file name ``_some_page_name``. This will evaluate to ` `Some Page Name`` with an extra space before Some.

I am perfectly fine with doing the one above! It looks a lot cleaner and I think that edge case wouldn't affect the pageTitle in the jsx file.

theboynextdoor: Sweet! Completely forgot about `.trim()`. Will slap it on with comments explaining the scenario I explained the examples you've mentioned.

kylemh:
Wow! Thanks so much for your contribution. It means the world to us here at Operation Code.

As a token of our gratitude, please fill out [this form](https://goo.gl/forms/r02wt0pBNhkxYciI3) to receive some stickers and a handwritten "thank you" note. ❤️

PR #274, opened 2018-11-13
Save docker images in ECR

wimo7083:
# Description of changes

This will start saving the docker images and build the k8s configs.


wimo7083:
The more I read about this, I think it's going to be better to get it into docker_hub.

kylemh:
Updated documentation and got ride of the node_js flag. The build is failing saying incompatible version of node. Why would it be using 8.9.3? I think something in CI is choosing it for us.

wimo7083:
Still deciding if I'm going to use ECR or dockerhub for all. Either way I'm waiting on some assistance with credentials.

wimo7083:
Per Kyle's request this is no longer needed.

ohaiwalt:
you closed it right about when I fixed the problem :shakes_fist:

ohaiwalt:
https://github.com/OperationCode/operationcode_infra/issues/64

PR #299, opened 2018-11-24
Use postcss.config.js inside Storybook

kylemh:
# Description of changes
We were defining PostCSS rulesets in two places which is an error-prone practice. Storybook _IS_ using a plugin not used by the application, but this change still ensures that both the app and storybook are getting their PostCSS rulesets from the same place.


kylemh:

Built with commit ee142f41f790daff963236a425ac521c73b89a56

https://deploy-preview-299--operation-code-storybook.netlify.com

Issue #10, opened 2018-07-16
Integrate Cypress.io


kylemh:
# Feature
## Blocked!
until #17 and #164 are accomplished, there will be too much churn to really justify the e2e tests that I'm hoping to create.

## Why is this feature being added?

[Cypress.io](https://cypress.io)
Because we want the best e2e and integration testing the industry has to offer 💪 

## What should your feature do?
TBA

takethefake:
would be interested in doing this when the time is ready!

kylemh:
Once #164 and #17 are completed, I can get going on setting up Cypress configuration so that we can begin to add e2e testing per page 💃 

Thanks for being ready to step up to the plate.

ariaaskarijob:
Would love to know when this is ready to contribute and collab!

kylemh:
Closed in #379 

Will add more issues for e2e test specs later!

PR #302, opened 2018-11-26
Parallel Lint/Test with Travis CI

### LINKED ISSUE #285: Parallelize linting and testing in our Travis build
# Feature request
Testing and linting doesn't have to run sequently. We could see a small build-time decrease if we parallelized those steps.

Use Travis build matrix or jobs to make linting and testing run in parallel.

RaevLogic:
# Description of changes
Introduced build stages in the travis.yml file so the linting and testing scripts would run in parallel.

# Issue Resolved

Fixes #285

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,10 +21,13 @@ before_script:
   - chmod +x ./cc-test-reporter
   - ./cc-test-reporter before-build
 
-script:
-  - yarn lint
-  - yarn test:ci
-  - yarn build
+jobs:
+  include:
+    - stage: lint/test
+      script: yarn lint
+      script: yarn test:ci
+    - stage: build
+      script: yarn build
 
 after_script:
   - greenkeeper-lockfile-upload


kylemh:

Built with commit 5b17465daa09ff93669f94865a27ec7be27601e4

https://deploy-preview-302--operation-code-storybook.netlify.com

RaevLogic:
That didn't work. `yarn lint` was never run. Also, according to the job log, the CI went through the `yarn test: ci` script, then went through all of the installation stuff again before moving on to the `yarn build` script. That actually added on to the build time rather than decreased it.

package.json
```diff
@@ -58,6 +59,7 @@
     "babel-loader": "^8.0.4",
     "babel-plugin-module-resolver": "^3.1.1",
     "classnames": "^2.2.6",
+    "concurrently": "^4.1.0",
```
kylemh: You can move this to devDeps cuz it's only needed on build-time

kylemh:
After a ton of discussion, we've come to realize that parallelization of lint/test either via concurrent node processes or via truly parallel build processes is a good idea. Truly parallel builds make our build time significantly larger, and concurrent node processes seem to only add a few seconds to the process.

When #10 is implemented, parallelized builds will DEFINITELY be helpful, so we'll leave comments here and consider #285 complete.

RaevLogic:
Once we integrate Cypress via #10, I would still keep the 'Concurrently' package in mind. Looking over the Travis build when Concurrently was implemented, the build time was about average at 3min 28sec and we were able to get Travis to run linting and testing in parallel, on one host/VM. Running Cypress may need its own VM, but I like to keep our options open.

PR #333, opened 2019-01-12
integrated postcss-custom-media

markchernov:
# Description of changes

Integrate PostCSS Media Variable
# Issue Resolved

Fixes #295

PR #344, opened 2019-02-21
Added teamjs and destructuring states

cmgonzales:
# Description of changes

# Issue Resolved

Fixes #


kylemh:

Built with commit edf4b2705f1424411296530edb8cf2a373186d32

https://deploy-preview-344--operation-code-storybook.netlify.com

pages/styles/teams.css
```diff
@@ -0,0 +1,24 @@
...
+
+.staffMembers {
+  padding: 20px 0;
+}
+
+.foundingMembers {
+  padding: 2rem 0 0 0;
+}
+
+@media screen and (max-width: 840px) {
```
kylemh: let's use one of our viewport dimensions here in a custom media query!

Check out this PR for examples: https://github.com/OperationCode/front-end/pull/333/files

pages/styles/teams.css
```diff
@@ -0,0 +1,24 @@
+.boardMembers {
+  align-items: flex-start;
```
kylemh: It looks like most of these rules get overridden at the bottom of this file.

flex-wrap is the only attribute here that actually applies to the DOM

pages/team.js
```diff
@@ -0,0 +1,95 @@
+// eslint-disable-next-line no-restricted-imports
```
kylemh: Can you remove this line and then delete the next line?

You don't actually need to import React on any file in the pages folder. That's done for you.

pages/team.js
```diff
@@ -0,0 +1,95 @@
...
+import React from 'react';
+import { getTeamMembersPromise } from 'common/constants/api';
+import { s3 } from 'common/constants/urls';
+import QuoteBanner from 'components/_common_/HeroBanner/HeroBanner';
+import Section from 'components/_common_/Section/Section';
+import TeamCard from 'components/_common_/TeamCard/TeamCard';
+import styles from './styles/team.css';
+
+export default class team extends React.Component {
+  constructor(props) {
```
kylemh: We don't need the constructor here.

Just do

```
state = {
  boardMembers: [],
  staffMembers: [],
}
```

pages/team.js
```diff
@@ -0,0 +1,95 @@
...
+    super(props);
+
+    this.state = {
+      boardMembers: null,
+      staffMembers: null,
+    };
+  }
+
+  async componentWillMount() {
```
kylemh: Data fetching should occur in cDM (`componentDidMount()`).

It _sounds_ like it should happen on cWM (`componentWillMount()`), but
- cWM is actually deprecated as of React 16.3
- An API call with fetch or axios inside componentWillMount will not return before the first render. This means the component will render with empty data at least once. Basically, cDM fetches guarantee only one rerender, but with cWM you could have 2.

kylemh:
The rendering issues have been resolved! All that's left is to polish the stylings and add an e2e test with Cypress.

PR #350, opened 2019-02-24
Pr/css3 to postcss

### LINKED ISSUE #293: Use postcss-custom-properties to statically represent CSS3 variables on build-time
# Feature request
CSS3 variables are great! We can avoid using a preprocessor while still doing enough to avoid hard-coding stuff in a CSS Module system. One problem is that CSS3 variables [don't have the best browser support](https://caniuse.com/#feat=css-variables).

We can use [postcss-css-variables](https://github.com/MadLittleMods/postcss-css-variables) to statically extract variables on build-time! Add the dependency and integrate it within `postcss.config.js`. Please uncomment `env` within `postcss.config.js` and only leverage library when `env === 'production'` ([see here for more on ctx](https://github.com/postcss/postcss-loader#context-ctx)).

```javascript
/* eslint-disable global-require */
module.exports = ({ file, options, env }) => {
  const ENV = env.toLowerCase();
  const isProd = ENV === 'production';

  return {
    ...options,
    plugins: [
      require('postcss-import')({ root: file.dirname }),
      isProd ? require('postcss-css-variables')() : false,
      require('autoprefixer')()
    ],
  };
};
```

juliantrueflynn:
# Description of changes

For production only, converts CSS3 variables with postcss for the benefit of wider browser support.
 
# Issue Resolved

Fixes #293


kylemh:

Built with commit abac9bf71efbe6e05eacb8d8e5fdb2a48e2f7627

https://deploy-preview-350--operation-code-storybook.netlify.com

PR #353, opened 2019-02-26
Pr/events page

### LINKED ISSUE #166: Create /events page
# Feature
## Why is this feature being added?

Our design team has come up with a beautiful redesign to simplify UX.

## What should your feature do?
Implement the newly-designed landing page via `pages/events.js`
Here is the mock-up: https://invis.io/KSO12EFRU7E#/323662230_OC_Events_Landing_Page

## Temporary Note
This project requires back-end labor to finish (creating events surrounding our meetup groups and events). In the meantime, you can act upon a JSON of you shape/choosing in order to mock the data represented in the design mock.

juliantrueflynn:
# Description of changes

Build /events page matching new mock design.

# Issue Resolved

Fixes #166


kylemh:

Built with commit 3ba7e148ca1934fd2f2365d01d293a53e4c47889

https://deploy-preview-353--operation-code-storybook.netlify.com

kylemh:

Built with commit 6f1eddd07541fa798e7a889c12382ec4c69ee44b

https://deploy-preview-353--operation-code-storybook.netlify.com

kylemh:
Great job! Looks nice on immediate view. I'll check this out deeper late tonight.

pages/events.js
```diff
@@ -0,0 +1,271 @@
...
+          {carouselItems.map(item => (
+            <div className={styles.eventsCarouselItem} key={item.label}>
+              {item.label}
+              {item.date && <div className={styles.eventsCarouselItemDate}>{item.date}</div>}
+            </div>
+          ))}
+        </Carousel>
+      </div> */}
+
+      <HeroBanner
```
kylemh: Let's use Content here when you've adjusted that component.

Hoping it will simplify styles and DOM too

```<Content ... columns={/* all the FlatCards */} />```

File: pages/events.js [hunk 2]
```diff
@@ -0,0 +1,271 @@
...
+        backgroundImageSource={`${s3}redesign/heroBanners/events.jpg`}
+        className={styles.meetupCardsBanner}
+      >
+        <div className={styles.meetupCardsRow}>
+          <div className={styles.col}>
+            <FlatCard className={styles.meetupCard}>
```
kylemh: While adjusting the FlatCard to allow for no image to be passed like we discussed on Slack, I wonder if you could also define an opinionated way for a LinkButton or Button to show up along the bottom bar 🤔 

These exact cards also appear on the code schools page for MooCs, so it'd be used in many places (although those FlatCards have images AND the bottom bar button). I've got ideas on how to give FlatCard that prop, but am interested in seeing how you'd solve it.

pages/styles/events.css
```diff
@@ -36,8 +16,14 @@
...
 }
 
-.meetupCard div > div:first-child {
-  display: none;
+.meetupCard > div > div:nth-child(2) {
+  margin-bottom: -3.6em;
+  text-align: center;
+}
+
+.meetupCard > div > div:nth-child(2) a:hover {
```
kylemh: Can you define these rules within the context of the FlatCard?

They'll always need to be this way. Not just on this page.


```
{Button && <div className={classNames(styles.flatCardButton, styles.alignCenter)}>{Button}</div>}

//////

.flatCardButton > button:focus,
.flatCardButton > button:hover,
.flatCardButton > a:focus,
.flatCardButton > a:hover {
  background-color: var(--white);
  color: var(--secondary);
}
```

pages/styles/events.css
```diff
@@ -36,8 +16,14 @@
   flex-direction: column;
 }
 
-.meetupCard div > div:first-child {
-  display: none;
+.meetupCard > div > div:nth-child(2) {
```
kylemh: You already defined `text-align: center` within the context of the FlatCard... So, if you need it here as well, there's something wrong on that end. Maybe apply it using `.flatCardButton > button` (etc) like I used below, instead.

The `margin-bottom: -3.6em` is exactly what I was hoping to get away from with putting the button into the context of the FlatCard.

I'm okay leaving this here for now, because it'll be a more complex styling, but I'd like to see if we can use CSS within the FlatCard to _always_ have the bottom Button/LinkButton show up where it is here (regardless of content).

kylemh:
Awesome work. Love all the deleted diff in the last two commits 🤤

pages/styles/events.css
juliantrueflynn: The `margin-bottom: -3.6em` (`-2em` for the padding, plus `-1.6em` to center) was to position the button over the bottom border. I wouldn't consider it "bad", and I'm not sure how else you would do this that wouldn't involve negative spacing. Dealing with `position: absolute` would seem like a bigger PITA and still require negative spacing.

We could reduce the amount of negative spacing necessary by eliminating the `2em` padding added by the `borderContainer` class.

For text-alignment on FlatCard button: you're right I missed that, and about to push update.

kylemh: You could do position relative off of the content of the card and have it be a constant number. I didn’t call it bad, but I would hope we don’t have to do it because it’s a bigger PITA to have to change styling based upon content.

juliantrueflynn: I should've expressed that better, didn't imply you were calling it bad but that doing it that way with negative spacing wouldn't be "incorrect".

I also should've just tested that out before opening my yapper. Turns out I had an unnecessary flex styling that I removed and using position absolute didn't require negative spacing for the button. I just pushed a change for moving those updates to FlatCard too while I was at it.

pages/styles/events.css
```diff
@@ -0,0 +1,132 @@
...
+  max-width: 500px;
+}
+
+.meetupCard {
+  padding: 0;
+}
+
+.meetupCard > div {
+  padding: 2rem;
+}
```
kylemh: Do you think these style overrides should be applied to all FlatCards too? Compare result against Success Story on homepage or the component itself on Storybook. I feel like we can deliver something that works for all usages without any style overrides.

PR #366, opened 2019-03-09
Pr/static custom server

juliantrueflynn:
# Description of changes

Setup custom server to allow serving selected files in `/static` from the root. As of now this is setup to only serve a default robots.txt (that allows scraping all pages) and favicon.ico. This was implemented following [Next.js's example](https://github.com/zeit/next.js/tree/canary/examples/root-static-files).

This PR does not implement @wimo7083 [suggestion](https://github.com/OperationCode/front-end/issues/238#issuecomment-439648148) of serving all files from a directory (ex: `/public`, it doesn't have to be in `/static` either), but that's possible if necessary.

**Note**: files in static that are selected to be served at root, still can be accessed through the static path. Both of these work: `/static/favicon.ico` and `/favicon.ico`.

# Issue Resolved

Fixes #238

PR #368, opened 2019-03-09
Revert "Pr/static custom server"

kylemh:
Reverts OperationCode/front-end#366

Getting this error in CI after merging: https://github.com/zeit/now-cli/blob/master/errors/verification-timeout.md

I didn't think hard enough about the initial PR... Our staging deploys aren't running the server - they're just static exports. We only found this issue after merge because that's when the server was actually consumed.


kylemh:

Built with commit 59273d5a70ee153dde2e953f2d08bd2c2dfadd5d

https://deploy-preview-368--operation-code-storybook.netlify.com

juliantrueflynn:
Do you think the npm script that is breaking this is `export`? Also, not sure of best way to recreate issue in order to fix this.

kylemh:
No, export should work regardless of server changes. I'll have to investigate, as I really have no idea.

PR #428, opened 2019-04-21
Remove cookie options

kylemh:
# Description of changes

We're using JWT, and storing cookies on the client. It's okay to store them insecurely, because we're only using them in headers in requests. It's up to the server to determine if they're valid.

Aside: Cypress uses Chai assertions which cause an ESLint error to go off. So, I've added an override of the rule for `*.spec.js` files.

# Issue Resolved

Hopefully fixes logging in and registration

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -113,6 +113,12 @@
       }
     },
     {
+      "files": ["*.spec.js"],
+      "rules": {
+        "no-unused-expressions": ["off"]
+      }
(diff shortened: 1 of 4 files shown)


kylemh:

Built with commit 61513e6546219ffd7177af5a3f9a57ef6bedd156

https://deploy-preview-428--operation-code-storybook.netlify.com

PR #440, opened 2019-04-26
login and registration "sad path" cypress tests

KrDimitrov:
# Description of changes

Implemented new cypress tests for login and registration forms, where a user is expected to mess something up.
# Issue Resolved

Fixes #431

PR #443, opened 2019-04-27
Catch potential error resulting from server being down

kylemh:
# Description of changes

- It's a valid case to fail gracefully from in our UI
- It's needed for #440 to pass
- Move messages to their own file
- Move mock generators to their own file


kylemh:

Built with commit b5852f543fc9715b2701c4264beeba599faff7c0

https://deploy-preview-443--operation-code-storybook.netlify.com

components/RegistrationForm/__tests__/RegistrationForm.test.js
```diff
@@ -171,26 +169,22 @@ describe('RegistrationForm', () => {
...
+    const existingUser = await mockUser('kylemh.email12@gmail.com');
 
     OperationCodeAPIMock.onPost('users', {
       user: {
-        email: testUser.email,
-        password: testUser.password,
-        first_name: testUser.firstName,
-        last_name: testUser.lastName,
-        zip: testUser.zipcode,
+        email: existingUser.email,
```
juliantrueflynn: This is where something like `hump`, a util, or decamelizing in backend would be helpful and make things cleaner for client. I could always create that util for decamelizing too since none of these values are nested and it would seem easy.

kylemh: You mean on the back-end? You're free to, but it's going alright for now!

PR #514, opened 2019-05-20
WIP: Back-end rewrite compatibility

AllenAnthes:
# Description of changes
PR created in order to preview integration with new back-end


kylemh:

Built with commit 7086ae83a258917d678e75235aa2657131ea8eec

https://deploy-preview-514--operation-code-storybook.netlify.com

common/constants/api.js
```diff
@@ -22,6 +22,12 @@ export const loginUser = ({ email, password }) =>
     password,
   }).then(({ data }) => data);
 
+export const passwordReset = ({ email }) =>
+  post('auth/password/reset/', { email }).then(({ data }) => data);
+
+export const passwordResetSubmit = values =>
+  post('auth/password_reset/confirm/', values).then(({ data }) => data);
```
kylemh: surprised this isn't `auth/password/reset/confirm`, but that's up to you!

pages/password_reset/confirm.js
```diff
@@ -0,0 +1,69 @@
...
+  state = {
+    success: false,
+  };
+
+  static async getInitialProps({ query: { uid, token } }) {
+    return { uid, token };
+  }
+
+  onSuccess = () => {
+    this.setState({ success: true });
```
kylemh: can you booleanify this variable name?

components/PasswordResetSubmitForm/PasswordResetSubmitForm.js
```diff
@@ -0,0 +1,119 @@
...
+      await passwordResetSubmit({ ...values, uid, token });
+      actions.setSubmitting(false);
+      actions.resetForm();
+
+      await onSuccess();
+    } catch (error) {
+      actions.setSubmitting(false);
+      if (error.response && error.response.status === 400) {
+        this.setState({
+          errorMessage: 'Could not reset password.  Reset token expired or invalid',
```
kylemh: You don't want to pass me an error message in the response?!?! 😢 

Then `getErrorMessage` would just handle it for ya!

pages/password_reset/confirm.js
```diff
@@ -0,0 +1,69 @@
...
+          <Link href="/login">
+            <a>Click here to Login</a>
+          </Link>
+        </div>
+      );
+    }
+    const { uid, token } = this.props;
+
+    if (!uid || !token) {
+      return <p>The provided credentials were either invalid or expired.</p>;
```
kylemh: Instead of a `p` element, it'd be super cool if you used an `Alert` component.

pages/password_reset/index.js
```diff
@@ -0,0 +1,36 @@
+import Head from 'components/head';
+import HeroBanner from 'components/HeroBanner/HeroBanner';
+import Content from 'components/Content/Content';
+import PasswordResetForm from 'components/PasswordResetForm/PasswordResetForm';
+import { passwordReset } from 'common/constants/api';
+
+class PasswordReset extends React.Component {
+  state = {
+    success: false,
```
kylemh: booleanify! ❤️

pages/password_reset/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+  onSuccess = () => {
+    this.setState({ success: true });
+  };
+
+  render() {
+    const { success } = this.state;
+
+    const content = success ? (
+      <p>A confirmation has been sent to the provided email address.</p>
```
kylemh: Alert here as well :) `type="success"`

kylemh: Also, I don't mind the earlier `getContent`, but I'm a big fan of throwing everything into render. With something as simple as a small ternary like this, I'd really prefer it there instead of defined as an extra variable.

pages/password_reset/index.js
```diff
@@ -3,31 +3,38 @@ import HeroBanner from 'components/HeroBanner/HeroBanner';
 import Content from 'components/Content/Content';
 import PasswordResetForm from 'components/PasswordResetForm/PasswordResetForm';
 import { passwordReset } from 'common/constants/api';
+import Alert from '../../components/Alert/Alert';
```
kylemh: Can you do the alias import path? `components/Alert/Alert`

.eslintrc.js
```diff
@@ -197,15 +197,16 @@ module.exports = {
       'error',
       {
         whitelist: {
+          ctx: true,
```
kylemh: thank you for alphabetizing ❤️

common/constants/api.js
```diff
@@ -17,14 +17,17 @@ export const createUser = ({ email, password, firstName, lastName, zipcode }) =>
...
-    email,
-    password,
-  }).then(({ data }) => data);
+  post(
+    'auth/login/',
+    {
+      email,
+      password,
+    },
+    { noAuth: true },
```
kylemh: what's `noAuth`?

common/utils/api-utils.js
```diff
@@ -16,22 +16,22 @@ export const get = async (path, { token } = {}) =>
     headers: setAuthorizationHeader(token),
   });
 
-export const post = async (path, body) => {
-  const result = await OperationCodeAPI.post(`/${path}`, body, {
-    headers: setAuthorizationHeader(),
+/**
+ * @param {string} path
+ * @param {{any}} body
+ * @param {{noAuth?: boolean}}options
```
kylemh: you'll wanna add a space between } and options

common/utils/api-utils.js
```diff
@@ -16,22 +16,22 @@ export const get = async (path, { token } = {}) =>
     headers: setAuthorizationHeader(token),
   });
 
-export const post = async (path, body) => {
-  const result = await OperationCodeAPI.post(`/${path}`, body, {
-    headers: setAuthorizationHeader(),
+/**
+ * @param {string} path
+ * @param {{any}} body
```
kylemh: `{?Object.<string,any>}` is prolly good here

common/utils/api-utils.js
```diff
@@ -16,22 +16,22 @@ export const get = async (path, { token } = {}) =>
     headers: setAuthorizationHeader(token),
   });
 
-export const post = async (path, body) => {
-  const result = await OperationCodeAPI.post(`/${path}`, body, {
-    headers: setAuthorizationHeader(),
+/**
```
kylemh: thanks for the JS Docs 😍

common/constants/api.js
kylemh: I see now. Maybe: `hasAuthHeaders` for naming.

At the same time, is it really necessary? What's wrong with always setting auth headers for every request?

components/SocialLoginGroup/SocialLoginGroup.js
```diff
@@ -0,0 +1,75 @@
...
+  static defaultProps = {
+    className: undefined,
+  };
+
+  state = {
+    errorMessage: '',
+  };
+
+  onFailure = () => {
+    this.setState({ errorMessage: "Couldn't log in with Google" });
```
kylemh: Nice pattern. What about Facebook?

components/SocialLoginGroup/SocialLoginGroup.js
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import { string, func } from 'prop-types';
+import FacebookLogin from 'react-facebook-login/dist/facebook-login-render-props';
+import classNames from 'classnames';
```
kylemh: nit: can you move classNames import above Facebook?

components/SocialLoginGroup/SocialLoginGroup.js
```diff
@@ -0,0 +1,75 @@
...
+              <button
+                type="button"
+                onClick={renderProps.onClick}
+                className={classNames(styles.loginButton, styles.facebookButton)}
+              >
+                Login with Facebook
+              </button>
+            )}
+          />
+        </div>
```
kylemh: Can you add a new line here? Reason: The blank line is a quick visual indicator that the fragment is necessary.

common/constants/api.js
AllenAnthes: Some funky Django stuff.  It appears that if a request has a JWT in the auth header the very first thing it does it check if it's valid, before it even gets to the view to see if it's an authenticated route or not.

I agree it's kinda weird.  I'm planning on looking into it some more, but it fixed an intermittent issue I was having with stale creds.

components/SocialLoginGroup/SocialLoginGroup.js
AllenAnthes: The Facebook component only takes the single callback.  Planning on figuring out how/where a facebook login failure comes in.

common/utils/api-utils.js
AllenAnthes: Man I would have sworn I fixed that.

pages/index.js
```diff
@@ -98,6 +99,16 @@ export default () => (
 
     <SponsorsSection />
 
-    <JoinSection />
+    {!isLoggedIn && <JoinSection />}
```
kylemh: note that there are other locations that <JoinSection> is rendered.

I wonder if it's best to connect that thing to redux so we don't have to think about it when we add it to other future pages... 🤔

common/constants/navigation.js
```diff
@@ -21,28 +21,29 @@ const eventsLink = {
   href: '/events',
 };
 
+export const loginLinks = {
```
kylemh: Ah now I see.

Why not simply make two arrays of navItems? One for loggedIn and one for loggedOut. Export/Import both and choose the right one based off the state variable at the implementation level.

components/Nav/NavMobile/NavMobile.js
```diff
@@ -59,6 +58,13 @@ export default class NavMobile extends Component {
                   </Link>
                 </li>
               ))}
+              {isLoggedIn && (
+                <li className={styles.li}>
+                  <button type="button" className={styles.button} onClick={logout}>
```
kylemh: Now I understand this.

I think we can get around this manipulation by making Logout a link that goes to `/login?loggedOut=true`

Then, on the login page, check for that query param in context and do the necessary unauth.

With that done:
- The abstraction pattern within the nav area is maintained.
- Every item in the nav is truly a link
- We can easily dynamically render a success Alert letting the user know they've logged out

AllenAnthes: Love it

components/SocialLoginGroup/SocialLoginButtons.js
```diff
@@ -0,0 +1,49 @@
...
+        >
+          Login with Facebook
+        </button>
+      )}
+    />
+  </>
+);
+
+SocialLoginButtons.propTypes = {
+  onSuccess: func.isRequired,
```
kylemh: Lame nit: can you put proptypes above the definition? So far in the repo proptypes are always at the top of the file.

kylemh:
Gonna merge this as it's growing. Further work can be done in small pull requests. I'm gonna focus on testing uncovered code for an hour today and most of tomorrow.

PR #492, opened 2019-05-13
Cypress Visual Regression Test Adjustments

kylemh:
# Description of changes
- Move visual regression tests to their own folder
- Create a cypress snapshot creating utility to ensure more consistent screenshots (and to do full page of mobile and desktop)
- Use arrow funcs in all `describe`
- Add a recurrently appearing file to `.gitignore`

# On Hold
We're going to add Docker to the equation so that testing headlessly is the same between CI and local environments. That should remove the variability between fullPage screenshots I'm seeing. On top of that, this PR would lead to much larger build times, so I'm going to put it on hold for #449 as well.

PR #544, opened 2019-05-27
Remove visual regression tests

kylemh:
# Description of changes
This removes our existing visual regression tests. I'm making this move for many reasons.

- As it stands, the tests take a screenshot of the top of the page. I'd argue that it's not providing much value in this manner, since regressions could exist below the fold. This makes these tests act as potential false-positives.
- Our build times are _massive_ now. CI takes upwards of 20 minutes on occasion, mostly because each screenshot diff takes about one minute.
- I was hoping to get #492 resolved, but it wasn't working out.

It may be best for us to simply wait for first-class visual regression diffing support from the Cypress team or at least until https://github.com/cypress-io/cypress/issues/2175 is resolved. Until then, I'm gonna speed up our build-time, so we can move fast and break things.

I've left in the dependencies and configuration work so that we don't have to do that again.


kylemh:

Built with commit 87c3caa8953307a354a4c45cb3799da45b2532e9

https://deploy-preview-544--operation-code-storybook.netlify.com

PR #604, opened 2019-06-13
No nums in name

### LINKED ISSUE #528: Prevent numbers in name fields of all forms with Yup validations
# Feature request
A user is able to enter non-alphabetical characters into fields that typically would **only** contain them.

## Proposed solution
Use `yup` on all relevant forms to prevent the entering of numbers into the aforementioned fields. Add e2e and unit test coverage on the matter.

osogrizz:
# Description of changes

Name fields on registration form no longer accept numbers or special characters. 
# Issue Resolved

Fixes #528


kylemh:

Built with commit b04efcd9cd82458e3b578003a0cc301de0cfc6f3

https://deploy-preview-604--operation-code-storybook.netlify.com

components/RegistrationForm/RegistrationForm.js
```diff
@@ -28,8 +28,12 @@ const registrationSchema = Yup.object().shape({
   'confirm-password': Yup.string()
     .required(validationErrorMessages.required)
     .oneOf([Yup.ref('password')], validationErrorMessages.passwordMatch),
-  firstName: Yup.string().required(validationErrorMessages.required),
-  lastName: Yup.string().required(validationErrorMessages.required),
+  firstName: Yup.string()
+    .required(validationErrorMessages.required)
+    .matches(/^[a-zA-Z]+$/, validationErrorMessages.name),
```
kylemh: So names _can_ have symbols.

Check this out https://www.regextester.com/93648

osogrizz: I was trying to figure out a good way to do this. I'm not too well versed in regex. Can I just do: exclude numbers and exclude special characters but allow some special character group? I'm assuming that's what the linked example does.

I'm thinking something like this maybe   `^[a-zA-Zà-žÀ-Ž \-]+$`

kylemh: Still missing apostrophes and commas. Consider copy-pasting the regex! That’s basically what I do at work 😂

kylemh: ```/^[a-z ,.'-]+$/i```

osogrizz: ``` /^[a-zà-ž ,'`-]+$/i ```  I had settled on this earlier as it allows northern wandering types or just names with accents and the like. But I can use your version as well.  


Screen Shot 2019-06-16 at 9 17 26 PM

PR #613, opened 2019-06-17
Hide logged out message after initial render 

### LINKED ISSUE #598: "Logged out successfully" alert should go away after /login re-renders
# Bug Report
If you logout, and then try to login again (but fail), the whole page re-renders, but the `loggedOut` query param still exists in the URL and the Alert still renders.

After logging out and attempting to log back in, but getting an error (or refreshing the page), the Alert should go away and the URL should read: `https://www.operationcode.org/login`

I logged out from a valid account, and then tried to login using Google but received an error.

### Operating System
MacOS latest

### Browser
Chrome latest

### Is This A Third-Party Issue?
No

### Screenshots
Screen Shot 2019-06-12 at 1 36 06 AM

AllenAnthes:
# Description of changes

Hides logged out alert after initial render by setting the alert message on initial mount, and removing it during any login action.

# Issue Resolved

Fixes #598


kylemh:

Built with commit 75b9e31fe274739600dd94da355e9cf2547e57ce

https://deploy-preview-613--operation-code-storybook.netlify.com

kylemh:
Every JS file in the pages folder gets bundled into now - even test files.

I’d recommend simply adding to the Cypress tests, since we don’t add the pages folder to test coverage. If you’re set on Jest, you’ll want to make a folder outside of pages for tests of those files.

cypress/integration/login.spec.js
```diff
@@ -90,3 +90,39 @@ describe('login', () => {
     cy.getCookies().should('have.length', 0);
   });
 });
+
+describe('login?loggedOut=True', () => {
+  beforeEach(() => {
```
kylemh: You can drop this `beforeEach` and `describe` and just put all of these tests in the top `describe`.

AllenAnthes: Good point.  When I started I didn't think I was going to use the same URL for each test.

Issue #664, opened 2019-07-13
Accessibility: Navigation submenus should only have 1 open at any time


jack-robs:
# Feature request
for keyboard accessibiity is when another submenu opens, if another is already open, close the previously opened one so that if someone using a screenreader tabs backwards, they're aware of the current context.

## Proposed solution
Unk - research/propose one w/ PR as well

## Potential alternative solutions
n/a

## Additional context
A11Y audit

danielasannino:
Will tackle this one.

PR #741, opened 2019-08-17
Screen Reader Accessibility

danielasannino:
# Description of changes

I added role and aria-label attributes to follow Web Accessibility Guidelines
# Issue Resolved

Fixes # Issue #664


kylemh:

Built with commit e20a21bfe07da38e6006ca92e4d8445ef23f90c8

https://deploy-preview-741--operation-code-storybook.netlify.com

components/Nav/Nav.js
```diff
@@ -56,7 +57,7 @@ export class Nav extends Component {
 
         <header className={styles.NavDesktop} data-testid="Desktop Nav">
           <div className={styles.navContainer}>
-            <nav>
+            <nav role='navigation'>
```
kylemh: See teh note below: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role#Description

```
Using the <nav> element will automatically communicate a section has a role of navigation. Developers should always prefer using the correct semantic HTML element over using ARIA.
```

components/Nav/Nav.js
```diff
@@ -67,7 +68,7 @@ export class Nav extends Component {
                 </a>
               </Link>
 
-              <ul className={styles.link}>
```
kylemh: This is unnecessary.

components/Nav/Nav.js
```diff
@@ -14,6 +14,7 @@ import NavMobile from 'components/Nav/NavMobile/NavMobile';
 import { hasValidAuthToken } from 'common/utils/cookie-utils';
 import styles from './Nav.css';
 
+
```
kylemh: Can you remove this?

components/Nav/NavMobile/NavMobile.js
```diff
@@ -34,7 +34,7 @@ export default class NavMobile extends Component {
     return (
       <header className={styles.NavMobile}>
         <Link href="/" prefetch={false}>
-          <button
+          <button role='button' aria-label='Mobile Navigation Button'
```
kylemh: I'm glad you touched this so I can see a new problem!

This should be `a` tag, not a button. After you change it, you can ditch the role, name, and type attributes from this element.

In terms of [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute), you want to describe the action the occurs when clicking on the link. Since this link navigates the user to the home page, `Go Home` seems like a good choice.

components/Nav/NavMobile/NavMobile.js
```diff
@@ -47,14 +47,14 @@ export default class NavMobile extends Component {
           </button>
         </Link>
 
-        <button
+        <button role='button' aria-label='Mobile Navigation Button'
```
kylemh: There's no need for `aria-label` here, since the button's purpose is described in plain text for screen readers (see line 56)

components/Nav/NavMobile/NavMobile.js
```diff
@@ -47,14 +47,14 @@ export default class NavMobile extends Component {
...
           className={classNames(styles.button, styles.hamburger)}
           onClick={openMenu}
           type="button"
           name="dropdown"
         >
           <ScreenReaderOnly>Open Menu</ScreenReaderOnly>
-          <HamburgerIcon className={styles.hamburgerIcon} />
+          <HamburgerIcon className={styles.hamburgerIcon} aria-label='Mobile Navigation Trigger'/>
```
kylemh: you can ditch this

kylemh:
Some of these changes were helpful, but #664 says:

```
... when another submenu opens, if another is already open, close the previously opened one so that if someone using a screenreader tabs backwards, they're aware of the current context.
```

This has to do with the `NavListItem` component. You wont need to futz with any aria attributes - it's just a matter of ensuring only one submenu is open at any one time.

See [this fly-out menu example](https://www.w3.org/WAI/tutorials/menus/flyout/#use-button-as-toggle) for context.

You can accomplish this by creating a method for use on `onKeyDown` of the actual sublink items (anchor tags). If it's the first or last sublink item and the tab key is struck, set `areSublinksVisible` to false with `this.setState`

components/Nav/NavMobile/NavMobile.js
```diff
@@ -34,17 +34,13 @@ export default class NavMobile extends Component {
     return (
       <header className={styles.NavMobile}>
         <Link href="/" prefetch={false}>
-          <button
-            className={classNames(styles.button, styles.logoButton)}
-            type="button"
-            name="dropdown"
-          >
+          <a aria-label="Go Home" className={classNames(styles.button, styles.logoButton)}>
```
kylemh: You know what... instead of `aria-label`, let's use the ScreenReaderOnly component.

components/Nav/NavListItem/NavListItem.js
```diff
@@ -46,6 +46,14 @@ export default class NavListItem extends Component {
 
     const hasSublinks = props.sublinks.length > 0;
 
+    function tabDown() {
```
kylemh: You defined this method as `tabDown` but didn't consume it anywhere.

Did you test to see if this works?

components/Nav/NavListItem/NavListItem.js
```diff
@@ -54,6 +62,7 @@ export default class NavListItem extends Component {
             onMouseEnter={this.showSublinks}
             onMouseLeave={this.hideSublinks}
             role="link"
+            onKeyDown={this.onKeyPressed}
```
kylemh: You didn't define `onKeyPressed` anywhere in this component.

components/Nav/NavListItem/NavListItem.js
```diff
@@ -29,6 +29,13 @@ export default class NavListItem extends Component {
     areSublinksVisible: false,
   };
 
+  constructor() {
```
kylemh: We're using [@babel/plugin-proposal-class-properties](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties) (hidden by the next preset for babel) which automatically binds methods to the class instance if you use arrow methods!

So ditch the whole constructor, and just do this in the future.

```javascript
areSublinksVisible = (event) => {
}
```

components/Nav/NavListItem/NavListItem.js
```diff
@@ -41,6 +48,19 @@ export default class NavListItem extends Component {
     this.setState(previousState => ({ areSublinksVisible: !previousState.areSublinksVisible }));
   };
 
+  areSublinksVisible(event) {
+    if (!this.submenu.contains(event.target)) {
```
kylemh: In general, the _only_ time you want to be referencing `this` in our application is to get access to `props`, `state`, or `setState`.

kylemh: What is `submenu` referring to in this situation?

components/Nav/NavMobile/NavMobile.js
```diff
@@ -34,17 +34,13 @@ export default class NavMobile extends Component {
     return (
       <header className={styles.NavMobile}>
         <Link href="/" prefetch={false}>
-          <button
-            className={classNames(styles.button, styles.logoButton)}
-            type="button"
-            name="dropdown"
-          >
+          <ScreenReaderOnly className={classNames(styles.button, styles.logoButton)}>
```
kylemh: This will make the image invisible. Search for other usages of `<ScreenReaderOnly>` to see how it should be used. You got rid of the anchor tag - you'll want to bring that back.

components/Nav/NavListItem/NavListItem.js
```diff
@@ -41,6 +48,19 @@ export default class NavListItem extends Component {
     this.setState(previousState => ({ areSublinksVisible: !previousState.areSublinksVisible }));
   };
 
+  areSublinksVisible(event) {
```
kylemh: I'm not sure I understand where you're going with this.

PR #789, opened 2019-09-21
Make every header a scroll link

### LINKED ISSUE #729: Make the `Heading` component act as scroll anchor
# Feature request
As a user, I'd love to be able to scroll to specific sections of our pages and link people to that exact section.

**Example:**
https://scrollmagic.io/examples/advanced/anchor_link_scrolling.html#middle

## Proposed solution
Make edits to the `Heading` component to render a link icon (use [this icon from FontAwesome](https://fontawesome.com/icons/link?style=solid)), but conditionally make it visible only on hover. Clicking it should change the route to include the hash of the scroll anchor and move the viewport such that the clicked header is at the top of the view.

walleyyang:
# Description of changes
Added HashLink component to Content component. This will enable the Content components that contains the title to have the link icon next to the title when hasHashLink is included. Also added HashLink component to other headings, but had to customize it. The link icon does go away when clicking the link icon or going directly to the link with the hashed URL.

Added HashLinks to the headings on these pages:
- Index
- About Us
- FAQ
- Podcast (interviewee names)
- Who We Serve
- Events
- Get Involved
- Sponsorship
- Leadership Circle

# Issue Resolved
Fixes #729

## Screenshots/GIFs
example


kylemh:

Built with commit 60ed715882604bce5e555b0394a6236df8549918

https://deploy-preview-789--operation-code-storybook.netlify.com

walleyyang:
> Lot of awesome stuff here.
> 
> Here are a few high-level comments to get you going on a round two.
> 
> Firstly, the biggest issue is that we have a lot of one-offs. IMO it's a red-flag that we've got `hashLinkIconOffset` inside themeMap and style variables because - taking a step back - this doesn't need to be used outside of a single component. By that I mean, let's edit the `Heading` component to include all the logic for HashLink. To make it a backwards compatible edit, provide a `headingLevel` prop which is just the number 1-6 to define `h1` - `h6` for the component.
> 
> Once that's done, you can go into all the one-offs (like in `Content`, `DonateSection`, `JoinSection`, etc) and use the `Heading` component and you can clean up the consumptions that show up like in `pages/get_involved.js`.
> 
> With this done, you should only have to change styles in just the `Heading` component, as opposed to all over the app.
> 
> Last three things:
> 
> 1. I'd love if we moved away from `enzyme` towards `@testing-library/react`. Will you edit your unit tests accordingly?
> 2. I think you can delete all the snapshot tests
> 3. I'd love to see some kind of integration test with Cypress asserting that clicking on a hashlink changes the URL and that navigating to a hash route works (you can test things via what exists in the viewport).

Done. Let me know if you need any of the snapshot tests deleted. I wasn't sure which ones you were talking about.

components/Heading/Heading.js
```diff
@@ -1,33 +1,188 @@
...
-      {children}
-    </h2>
-  );
+class Heading extends Component {
+  static propTypes = {
+    className: string,
+    id: oneOfType([string, number]), // reference for scroll anchors
+    anchorId: string,
+    customAnchorClass: string,
+    children: node.isRequired,
```
kylemh: If you don't make this children and just make it a prop for a string to render (like `text`), you can dynamically generate the `anchorId`! Two birds with one prop.

components/Heading/__tests__/Heading.test.js
```diff
@@ -16,16 +20,248 @@ describe('Heading', () => {
       </Heading>,
     );
   });
+});
+
+describe('Heading anchor and hashlink', () => {
```
kylemh: I appreciate that you went after total coverage, but keep in mind we have to maintain tests and we want to write them to instill confidence in our code (by protecting against regressions and asserting complex things work as we expect).

Keeping that in mind, we can drastically reduce the amount of tests here.

I'd say have:
- One test that the component renders at all with required props (snapshot test is fine here)
- Two tests that assert the link's [visibility/invisibility](https://github.com/testing-library/jest-dom#tobevisible)

and that's it. You don't want to get caught testing against class names or asserting that props get passed through. You want to think about it from the perspective of the user interacting with it. They just care that the heading component works (renders) and functions properly.

Testing it's functionality is more valuable in a browser-like env, like Cypress because you can assert that the scroll position gets changed and the URL gets changed. The only functionality you care about in Jest is the anchor being visible after `onMouseEnter` or when focused **AND** invisible after `onMouseLeave` and when not focused.

cypress/integration/hashlink.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    { pathName: 'leadership_circle', pathValue: '/leadership_circle' },
+    { pathName: 'code_schools', pathValue: '/code_schools' },
+    { pathName: 'jobs', pathValue: '/jobs' },
+    { pathName: 'press', pathValue: '/press' },
+    { pathName: 'branding', pathValue: '/branding' },
+    { pathName: 'team', pathValue: '/team' },
+  ];
+
+  paths.forEach(path => {
+    const test = `checks hashlinks on ${path.pathName} page`;
```
kylemh: Why make a variable for this when you can just define it inline?

cypress/integration/hashlink.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    { pathName: 'sponsorship', pathValue: '/sponsorship' },
+    { pathName: 'leadership_circle', pathValue: '/leadership_circle' },
+    { pathName: 'code_schools', pathValue: '/code_schools' },
+    { pathName: 'jobs', pathValue: '/jobs' },
+    { pathName: 'press', pathValue: '/press' },
+    { pathName: 'branding', pathValue: '/branding' },
+    { pathName: 'team', pathValue: '/team' },
+  ];
+
+  paths.forEach(path => {
```
kylemh: Nice move with a programatic test here!

cypress/integration/hashlink.spec.js
```diff
@@ -0,0 +1,26 @@
+describe('hashlink', () => {
+  const paths = [
```
kylemh: maybe `pagesWithHashLink` is a better variable name here?

cypress/support/commands.js
```diff
@@ -52,6 +52,20 @@ Cypress.Commands.add('setResolution', size => {
   }
 });
 
+Cypress.Commands.add('verifyHashLink', path => {
```
kylemh: I like that you did you research to the point of learning about custom commands, but they are more ideal when they're used across multiple test files. I'd say you can just put this logic into `hashLink.spec.js`

components/Heading/Heading.js
```diff
@@ -1,33 +1,188 @@
...
+
+  toggleVisible = display => {
+    this.setState({ isLinkIconVisible: display });
+  };
+
+  getHeading = () => {
+    const { props } = this;
+    const classes = this.getHeadingClasses();
+
+    switch (props.headingLevel) {
```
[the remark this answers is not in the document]
walleyyang: Nice. Thanks for the info.

components/Heading/Heading.js
```diff
@@ -1,33 +1,188 @@
...
+    if (theme === 'white' || theme === 'gray') {
+      visibleIcon += ` ${styles.iconFillBlue}`;
+    } else {
+      visibleIcon += ` ${styles.iconFillWhite}`;
+    }
+
+    return visibleIcon;
+  };
+
+  toggleVisible = display => {
```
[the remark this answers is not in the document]
walleyyang: The issue mentioned to make it visible on hover. Did you always want it visible and not just on hover?

components/Heading/Heading.js
```diff
@@ -1,33 +1,188 @@
...
+
+  state = {
+    isLinkIconVisible: false,
+  };
+
+  getAnchorId = () => {
+    const { props } = this;
+
+    return props.anchorId
+      .replace(/\s+/g, '-')
```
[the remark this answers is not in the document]
walleyyang: Got it. Didn't think too much about it since it's only for a few characters.

cypress/integration/hashlink.spec.js
walleyyang: That's a good point.

components/Heading/Heading.js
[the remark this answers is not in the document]
kylemh: Sorry - it looks you do always have it rendered, but note that `visibility: hidden` also hides the content for screen readers. Note that there exist many web users that do not use a mouse - sometimes due to a disability. Since you only make it visible `onMouseEnter`, it would be considered "inaccessible".

Note the styles that we use for the `ScreenReaderOnly` component.

```css
.ScreenReaderOnly {
  border: 0;
  clip: rect(0, 0, 0, 0);
  height: 1px;
  margin: -1px;
  overflow: hidden;
  padding: 0;
  position: absolute;
  width: 1px;
}
```

This will visibly hide the content, but still make it keyboard accessible. You could also use `onFocus` then for people who _can_ see, but choose to use the keyboard.

components/Heading/Heading.js
walleyyang: Nice! Thanks for pointing that out.

components/Heading/Heading.js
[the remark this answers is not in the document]
walleyyang: Does this work?
```
<ScreenReaderOnly>
  {isLinkIconVisible ? 'Hashlink icon visible' : 'Hashlink icon hidden'}
</ScreenReaderOnly>
<LinkIcon className={isLinkIconVisible ? visibleIcon : hiddenIcon} />
```

kylemh:
#### 📝Changed routes:


| [**/about**](https://operation-code-git-729-hash-link.operation-code.now.sh/about) | [**/events**](https://operation-code-git-729-hash-link.operation-code.now.sh/events) |
|:-:|:-:|
|<a href="https://operation-code-git-729-hash-link.operation-code.now.sh/about">Screenshot of /about</a><br /><sup><a href="https://microlink-cdn.s3.amazonaws.com/s/wVSqE34iDMAY9m/w5BWT7PqU59Nk_">(view full size)</a> |<a href="https://operation-code-git-729-hash-link.operation-code.now.sh/events">Screenshot of /events</a><br /><sup><a href="https://microlink-cdn.s3.amazonaws.com/s/ovkZn5Znk4qZ_r/LNNbmBU5JqDWw_">(view full size)</a> |



| [**/get_involved**](https://operation-code-git-729-hash-link.operation-code.now.sh/get_involved) | [**/**](https://operation-code-git-729-hash-link.operation-code.now.sh/) |
|:-:|:-:|
|<a href="https://operation-code-git-729-hash-link.operation-code.now.sh/get_involved">Screenshot of /get_involved</a><br /><sup><a href="https://microlink-cdn.s3.amazonaws.com/s/sdGm0Ep_yHnU0i/btidUAQZUJtKQ_">(view full size)</a> |<a href="https://operation-code-git-729-hash-link.operation-code.now.sh/">Screenshot of /</a><br /><sup><a href="https://microlink-cdn.s3.amazonaws.com/s/g8SDQbvrGZnlzz/y7hrZBxgrD3mg_">(view full size)</a> |



| [**/podcast**](https://operation-code-git-729-hash-link.operation-code.now.sh/podcast) |
|:-:|
|<a href="https://operation-code-git-729-hash-link.operation-code.now.sh/podcast">Screenshot of /podcast</a><br /><sup><a href="https://microlink-cdn.s3.amazonaws.com/s/tiSpTryxuTPJui/n1ERYRuNCB3wc_">(view full size)</a> |







Commit eb97147a486cfc98cc2bc5f29a29fbce9ae29195 (https://operation-code-git-729-hash-link.operation-code.now.sh).

components/Heading/Heading.js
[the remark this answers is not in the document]
kylemh: I didn't mean to imply you need to use `ScreenReaderOnly` component, but that you should copy it's styles for the `hiddenIcon` class you created.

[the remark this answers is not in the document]
kylemh: Also, there's a bonus round if you're interested!

I really don't care too much about this point, buuuuuuuuuuuuut...

Keep in mind there's a `:focus` and `:hover` in CSS, so you can do the visibility / invisibility with just CSS, making this component even simpler.

components/Heading/Heading.js
```diff
@@ -8,9 +8,8 @@ class Heading extends Component {
   static propTypes = {
     className: string,
     id: oneOfType([string, number]), // reference for scroll anchors
-    anchorId: string,
     customAnchorClass: string,
-    children: node.isRequired,
+    children: string.isRequired,
```
kylemh: You'll want to make this _not_ `children` and call it something else, so people don't accidentally pass JSX and break your dynamic anchorId generation.

walleyyang: The props previously contained anchorId and you mentioned that we can remove that by making the children a string. If I were to rename it something else then I would need to add back another prop like anchorId since it's currently just taking the text as the id. Did you want it reverted?

components/Heading/Heading.js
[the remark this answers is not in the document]
walleyyang: Adding the ScreenReaderOnly style to the link hides it, but how do you bring it back on hover? `.hashLinkContainer a {
  /* visibility: hidden; */
  border: 0;
  clip: rect(0, 0, 0, 0);
  height: 1px;
  margin: -1px;
  overflow: hidden;
  padding: 0;
  position: absolute;
  width: 1px;
}

.hashLinkContainer:hover a {
  visibility: visible;
}`

components/Heading/Heading.js
kylemh: children is a reserved prop name for React components. Just because we’ve listed it’s prop type as string doesn’t mean people will adhere to it. In that same message, I suggested renaming your `children` prop to `text` does that make sense?

Then you would be doing `<Heading text=“Who We Serve” headingLevel={1} />`

components/Heading/Heading.js
```diff
@@ -1,33 +1,188 @@
...
+            {props.children}
+          </h6>
+        );
+    }
+  };
+
+  getHeadingClasses = () => {
+    const { props } = this;
+    const customHeadingClass =
+      props.className === 'headingOurMission' ? styles.headingOurMission : styles.Heading;
```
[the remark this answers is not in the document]
walleyyang: Thanks for the tip! That made a huge difference.

components/Heading/__tests__/Heading.test.js
walleyyang: I ended up removing all the tests except for the Cypress ones. The icon is based on hover now and there is no more state to check visibility.

components/Heading/Heading.css
```diff
@@ -18,26 +32,58 @@
   margin: 0 10px 5px 10px;
 }
 
-.secondary {
```
kylemh: You wont want to delete the class names for the theme! Also, can you tie `fill` for the icon class names to CSS variables?

components/Heading/Heading.js
```diff
@@ -1,33 +1,67 @@
...
+    hasHashLink: true,
+    headingLevel: 2,
+    theme: 'secondary',
+  };
 
-function Heading({ children, className, hasHeadingLines, id, theme }) {
-  const classes = classNames(className, styles.Heading, styles[theme], {
-    [`${styles.headingLines}`]: hasHeadingLines,
-  });
+  render() {
```
kylemh: dang look how simple this got 😍

components/Heading/Heading.js
```diff
@@ -1,33 +1,67 @@
...
+    const { props } = this;
+    const anchorId = kebabCase(props.text);
+    const classes = classNames(styles[props.theme], {
+      [`${styles.Heading} ${props.className}`]: !props.hasHeadingLines,
+      [`${styles.headingOurMission}`]: props.className === 'headingOurMission',
```
kylemh: You'll still wanna kill this line and pass `headingOurMission` into that specific usage of `<Heading>` instead of in this component.

components/Heading/Heading.js
```diff
@@ -1,33 +1,67 @@
...
-    </h2>
-  );
+    return (
+      <div className={`${styles.headingContainer}`}>
+        {props.hasHashLink && (
+          <span
+            id={anchorId}
+            className={classNames(styles.anchor, {
+              [styles.anchorMargin]: props.customAnchorClass === 'anchorMargin',
+              [styles.anchorMarginPodcast]: props.customAnchorClass === 'podcast',
```
kylemh: You'll wanna kill this one too

components/Heading/Heading.js
```diff
@@ -1,33 +1,67 @@
...
+        )}
+        {props.hasHashLink && (
+          <div className={`${styles.hashLinkContainer}`}>
+            <a href={`#${anchorId}`} data-testid="Hash Link">
+              <LinkIcon className={styles.icon} />
+              <ScreenReaderOnly>Scroll Link</ScreenReaderOnly>
+            </a>
+          </div>
+        )}
+        <HeadingElement className={classes}>{props.text}</HeadingElement>
```
kylemh: Can you nest the HeadingElement inside the anchor tag? It's good that we're letting DOM-focused individuals (keyboard users) know it's a scroll link, but we also need to give context of what it's scrolling to!

```
<div className={`${styles.hashLinkContainer}`}>
  <a href={`#${anchorId}`} data-testid="Hash Link">
    <LinkIcon className={styles.icon} />
    <ScreenReaderOnly>Scroll Link</ScreenReaderOnly>
    <HeadingElement className={classes}>{props.text}</HeadingElement>
  </a>
</div>
```

components/Heading/Heading.js
```diff
@@ -1,33 +1,78 @@
...
+          <HeadingElement
+            className={classNames(
+              styles[props.theme],
+              styles.headingOurMission,
+              styles.headingLines,
+            )}
+          >
+            {props.text}
+          </HeadingElement>
+        )}
```
kylemh: Can you delete this?

components/Heading/Heading.js
```diff
@@ -1,33 +1,78 @@
...
+    const HeadingElement = `h${props.headingLevel}`;
 
-  return (
-    <h2 className={classes} id={id}>
-      {children}
-    </h2>
-  );
+    return (
+      <div className={`${styles.headingContainer}`}>
+        {props.hasHashLink && (
```
kylemh: I think you only meant to wrap the anchor with this!

components/Heading/Heading.js
```diff
@@ -1,33 +1,78 @@
...
+          <>
+            <span
+              id={anchorId}
+              className={classNames(styles.anchorSpan, styles[props.customAnchorClass])}
+            />
+            <div className={`${styles.hashLinkContainer}`}>
+              <a href={`#${anchorId}`} data-testid="Hash Link">
+                <ScreenReaderOnly>Scroll Link</ScreenReaderOnly>
```
kylemh:
```suggestion
                <ScreenReaderOnly>Scroll link for {text}</ScreenReaderOnly>
```

pages/podcast.js
```diff
@@ -76,7 +77,12 @@ class Podcast extends React.Component {
 
                   return (
                     <Card data-testid="Podcast Card" className={styles.podcastCard} key={name}>
-                      <h3>{interviewee}</h3>
+                      <Heading
+                        text={interviewee}
+                        theme="white"
+                        customAnchorClass="anchorMarginPodcast"
```
kylemh: What's the point of this?

components/Heading/Heading.css
```diff
@@ -19,25 +33,79 @@
 }
 
 .secondary {
-  color: var(--secondary);
+  background-color: var(--secondary);
```
kylemh: Don't define `background-color` on any of these, and you won't need `.transparentWhite` defined.

Then you can ditch the iconFill classes... In fact, I think I mislead you here. We should probably just drop theme-handling styles in this component. We should define `fill` instead the theme styles of the Container component.

pages/podcast.js
walleyyang: The podcast page requires a different top margin otherwise it'll look cropped at the top.

components/Heading/Heading.css
walleyyang: I'm not sure where you are going with this. Where do you want to handle passing in themes then? I needed transparentWhite for the Donate section with the flags in the background.

components/Heading/Heading.js
walleyyang: The Our Mission heading in the home page is a one off. How did you want to handle this?

components/Heading/Heading.js
walleyyang: I was having issues with wrapping this in the anchor tag. For example, my issue was that the link icon's right margin looked like it increased once the screen is at a small view port to where the heading wrapped into another line. Right now they are siblings and works. Hovering over the heading now displays the link icon.
hoverheadingtextexample

The current minor issue is the heading text wrapping again. However, this time its the additional line that is not centering. The heading current offsets the right margin so the text appears centered without the icon displaying. It looks fine with a single line. It has the centering issue with the wrapped line. I'm not sure how to fix this one.
example2
multiline

components/Heading/Heading.css
kylemh: The only reason `Heading` takes `theme` as a prop is because it's passed through by `Container`.

It only needs access to theme to define the header's text color, right?

components/Heading/Heading.js
```diff
@@ -54,7 +52,17 @@ class Heading extends Component {
...
                 )}
               >
-                {props.text}
+                <div className={`${styles.hashLink}`}>
+                  <a id={`${anchorId}-link`} href={`#${anchorId}`} data-testid="Hash Link">
+                    <ScreenReaderOnly>Scroll Link</ScreenReaderOnly>
+                    <LinkIcon className={styles.icon} />
+                  </a>
+                  {props.hasTitleUnderline ? (
+                    <span className={`${styles.underline}`}>{props.text}</span>
```
kylemh: You won't need string template literals here!

components/Heading/Heading.js
```diff
@@ -1,33 +1,84 @@
...
+          <HeadingElement
+            className={classNames(
+              styles[props.theme],
+              styles[props.className],
+              styles.headingLines,
+            )}
+          >
+            {props.text}
+          </HeadingElement>
+        )}
```
kylemh: I think you can just drop this and put `[props.hasHeadingLines]: styles.headingLines` in the `classNames` function call for the other `<HeadingElement>`

components/Heading/Heading.js
walleyyang: Updated this to handle "hasHeadingLines" and passed in the classname to set. This removes the condition to check specifically for headingOurMission.

components/Heading/Heading.js
walleyyang: The other one is for the HeadingElement that includes the link icon. They are structured differently. The one above it is for the hash link. It needs the span to keep the icon displaying after clicking it. It also requires the link to be wrapped in a div so it can initially be hidden unless hovered over. It also has the link and heading text as children so it won't cause the link icon to appear too far to the left of the heading link when the heading text wraps to another line.

kylemh: If I pass `hasHeadingLines` to a heading component, then it won't have a scroll link.

kylemh:
@walleyyang I went ahead and implemented what I was seeking, cuz I didn't want you to bang your head against something so big for all of Hacktoberfest. You did a great job - my changes were mostly moving things around!

kylemh:
I noticed there was only one usage of `Heading` with `hasHeadingLines` so I made an executive call and just deleted it.

PR #808, opened 2019-09-28
Bugfix(es) for mobile content

manthonyg:
# Description of changes
1. Made dropdown nav scrollable
2. Fixed horizontal scroll on index page
3. Fixed the way school cards rendered on small screens

### **problem:**
navscroll-theproblem

### **fix**
navscroll-thefix

### **problem**
horizscroll-theproblem

**culprit was in the sponsors section. before the images scroll into the viewport and are loaded, the image box was overflowing the viewport based on the length of the alt text. anything > 12 characters and it would cause the above issue.**

horizscroll-theproblem

### **fix**
horizscroll-thefix

### **problem**
codeschool-theproblem

### **fix**
using a `vw` width on small screens instead of `100%` 
codeschools-thefix

PR #822, opened 2019-10-03
Nav scroll 

### LINKED ISSUE #726: Mobile navigation is not scrollable
# Bug Report
When on mobile, a hamburger menu is shown on the page header which - when clicked - reveals the **mobile navigation** (`NavMobile`). That view is not able to scroll; not all items are visible.

When a mobile user attempts to scroll within the `NavMobile` it should scroll the list of navigation links - not the rest of the app. If impossible, it's still more desirable to scroll both.

Noticed a difference in the number of mobile nav links rendered on an iPhone 6 vs. iPhone X

manthonyg:
# Description of changes
The last PR (https://github.com/OperationCode/front-end/pull/808) addressed some of the mobile content issues and also made the mobile nav dropdown scrollable. The problem was that the body content was not locking with the nav dropdown open, so once it reached the end of the y-scroll for the nav it would then begin to scroll the body... no bueno for UX.

The idea of locking the body scroll for a modal or nav has actually been plaguing people for a while. There even exists a npm package (with a lot of weekly downloads) that addresses this issue.
https://www.npmjs.com/package/body-scroll-lock

I definitely messed with this for quite a bit, and I think this will work well as a hotfix that may need improvement eventually- but maybe not. It just feels like a less than eloquent way to go about it, but it accomplishes the goal. I would be happy to explore other options based on what OC team would like to do.

It should be noted that in some instances on mobile, if the user very deliberately double taps the background (what would appear behind the nav dropdown) it becomes scrollable. Double click again? The focus is once again on the nav dropdown with the body locked. Why? I dunno. It seems like it would not be a huge issue as I had to really try and make it happen, but I wanted to point it out.

Lastly, the `padding-bottom: 5rem` just helps boost the last nav item above the mobile browsers navigation.

# Issue Resolved
Fixed #726 

### Chrome

navscroll-chrome

### Safari

navscroll-safari


kylemh:

Built with commit e8dd6edcf875de4a1780cce7057cbcb8e49a5410

https://deploy-preview-822--operation-code-storybook.netlify.com

PR #842, opened 2019-10-05
feat: prevent form submission double click race condition

### LINKED ISSUE #521: Improve form submission user experience
# Feature request
When submitting the login and registration forms, as well as each step in the profile update form, we've used Formik's `setSubmitting` as a way to manage the state of the submit button being disabled or not. It seems, however, that the button will undisable _before_ the post-submit route transitions start, giving an opportunity for multiple submissions.

I'd like for it to stay disabled until the server responds _AND_ until the route changes.

## Proposed solution
None

## Additional context
Feel free to discuss here since there isn't a crystal clear acceptance criteria.

RanadeepPolavarapu:
# Description of changes

Set the Formik `setSubmitting` to `false` only after the route change has gone through. This prevents users from double click prior to the route change.

# Issue Resolved

Fixes #521


kylemh:

Built with commit 84b5cc8fd5473b6649b7c7aad0f23e68cfdd5c2b

https://deploy-preview-842--operation-code-storybook.netlify.com

kylemh:
#### 📝Changed routes:


| [**/join**](https://operation-code-git-fork-ranadeeppolavarapu-fix-521.operation-code.now.sh/join) | [**/login**](https://operation-code-git-fork-ranadeeppolavarapu-fix-521.operation-code.now.sh/login) |
|:-:|:-:|
|<a href="https://operation-code-git-fork-ranadeeppolavarapu-fix-521.operation-code.now.sh/join">Screenshot of /join</a><br /><sup><a href="https://microlink-cdn.s3.amazonaws.com/s/5kR2t06tf_MwaR/SOLZrYArHph7c_">(view full size)</a> |<a href="https://operation-code-git-fork-ranadeeppolavarapu-fix-521.operation-code.now.sh/login">Screenshot of /login</a><br /><sup><a href="https://microlink-cdn.s3.amazonaws.com/s/uF83ve5a7OKvcH/5CzK_xvFPmxW4_">(view full size)</a> |







Commit 84b5cc8fd5473b6649b7c7aad0f23e68cfdd5c2b (https://operation-code-git-fork-ranadeeppolavarapu-fix-521.operation-code.now.sh).

common/utils/auth-utils.js
```diff
@@ -4,7 +4,7 @@ import { setAuthCookies, removeAuthCookies, hasValidAuthToken } from './cookie-u
 
 export const login = ({ token, user }, routeTo = '/profile') => {
   setAuthCookies({ token, user });
-  Router.push(routeTo);
+  return Router.push(routeTo);
```
RanadeepPolavarapu: This yields a `Promise<boolean>` to let you know when the route change has successfully gone through.

kylemh: Ah. So then instead of return, when if we make this method async and then make this line the `await`?

(Same with join and login pages)

components/LoginForm/LoginForm.js
```diff
@@ -49,9 +49,12 @@ class LoginForm extends Component {
 
     try {
       const { user, token } = await login(values);
-      actions.setSubmitting(false);
-      actions.resetForm();
-      await onSuccess({ user, token });
+      const isRouted = await onSuccess({ token, user });
```
kylemh: If the onSuccess handler includes a route change, won’t this form unmount?!

common/utils/auth-utils.js
RanadeepPolavarapu: Yes, that would produce the same result.

components/LoginForm/LoginForm.js
RanadeepPolavarapu: The `onSuccess` handler does include a route change currently, prior to this change even.

RanadeepPolavarapu:
I'm a bit confused about *codeclimate/diff-coverage*. I added two test cases for both Login and RegistrationForm.

kylemh:
Don’t worry about that check. It’s designed to keep reviewers on their toes. In my opinion, coverage metrics are a guide, not a hard rule.

I’ve got confidence in this change, especially given the tests you’ve written.

components/LoginForm/LoginForm.js
kylemh: Right, but I’m saying you’ve re-ordered things and the other methods won’t call because the form will unmount.

common/utils/auth-utils.js
kylemh: Would you make that change? I like the explicitness of it.

RanadeepPolavarapu: Sure, I can do that.

components/LoginForm/LoginForm.js
RanadeepPolavarapu: I see. Is it necessarily a bad thing if the form will unmount? 

If we don't do this then we have the same issue with the double clicking.

kylemh: The only issue that I see is that the form wont reset in state on a route change and will be permanently set to submitting, but I can't think about any negative consequences. It's possible the form's state resets on unmount, in which case we could just ditch both methods anyways 🤷‍♂. Can you verify that logging in with the code that you currently have yields no run-time errors in the console?

RanadeepPolavarapu: Sure, I have just now verified with Network Throttling feature on Chrome (Slow 3G) that this change works as expected and that there are no errors in the console during runtime.

PR #852, opened 2019-10-06
add Pagination component

### LINKED ISSUE #447: Create pagination component
# Feature request
We will eventually be creating a page for our resources API. In preparation, we can accomplish some work ahead of time before the API is ready to go.

Create a pagination component to navigate multiple pages of results.

Here is the mock:
Screen Shot 2019-04-27 at 9 33 10 PM

**Adjustment**: Instead of coloring the active page blue, please wrap it in the same blue as a square border (for accessibility concerns).

**Note**:
I highly recommend building the component using Storybook as a work-bench and keeping track of the following edge cases:

- Not having arrows when one page exists
- Not having a left arrow visible when on page one
- Not having the right arrow visible when on the last of all possible pages
- Only displaying 9 numbers at a time (what happens when your active page is 15 of 30?)

prmsrswt:
# Description of changes
created `Pagination` component to navigate multiple Pages.

# Issue Resolved

Fixes #447 

## Screenshots/GIFs
image


kylemh:

Built with commit 921584ff9537be9031ff0786f7bf7aa1e14b6a38

https://deploy-preview-852--operation-code-storybook.netlify.com

components/Pagination/Pagination.js
```diff
@@ -0,0 +1,64 @@
...
+            <PaginationItem key="seperatorEnd" value="&hellip;" isClickable={false} />
+            <PaginationItem key={totalPages} value={totalPages} />
+          </>
+        )}
+      </>
+    );
+  };
+
+  return (
+    <div className={classNames(styles.Pagination, className)}>
```
kylemh: Can you make this a `nav` element?

components/Pagination/Pagination.js
```diff
@@ -0,0 +1,64 @@
...
+
+    const PaginationItemArray = [...Array(paginationLength)].map((_, index) => {
+      const page = index + paginationStart;
+      const isCurrent = page === currentPage;
+
+      return <PaginationItem key={page} value={page} isCurrent={isCurrent} />;
+    });
+
+    return (
+      <>
```
kylemh: Can you make this an `ol` element and make the `PaginationItem` component a `li` element?

You'll have to do some style overrides, but it'll be really semantic HTML 💃

components/Pagination/Pagination.js
prmsrswt: Okay I will do that

components/Pagination/Pagination.js
prmsrswt: Thanks for your input. Yeah I saw the tests, it reduced the complexity but this time another test failed because I forgot to update the storybook snapshot. 
Also I have added some tests for `PaginationItem`.

prmsrswt:
@kylemh I have created tests for both `Pagination` and `PaginationItem` and used `nav`, `ol` and `li` to keep the HTML semantic. Is there anything else I need to do?

kylemh:
I’ll get that review to within 24 hours.

prashant-kush:
No hurries, thanks!

components/Pagination/Pagination.js
```diff
@@ -0,0 +1,77 @@
...
+  currentPage: number.isRequired,
+  totalPages: number.isRequired,
+};
+
+Pagination.defaultProps = {
+  className: undefined,
+};
+
+// eslint-disable-next-line react/prop-types
+const PaginationItems = ({ currentPage, totalPages }) => {
```
kylemh: I would love a dev-only error that highlights when current page is > total pages (impossible)

components/Pagination/Pagination.js
```diff
@@ -0,0 +1,77 @@
...
+
+    return <PaginationItem key={page} value={`${page}`} isCurrent={isCurrent} testId={`${page}`} />;
+  });
+
+  return (
+    <>
+      {shouldTruncateStart && (
+        <>
+          <PaginationItem key="1" value="1" testId="1" />
+          <PaginationItem
```
kylemh: Can we just style an ellipse and render it here instead of forcing it into the PaginationItem component API? If you do that, you can drop `isClickable`

components/Pagination/Pagination.js
prmsrswt: Yup, also I guess it would be a good idea to check if `currentPage` is less than or equal to `0`.

components/Pagination/Pagination.js
prmsrswt: I used the `PaginationItem` component for ellipsis so it would retain visual consistency. So for example if we would change CSS properties like font-size, background color etc. of `PaginationItem` it would be applied on the ellipsis as well.

prmsrswt:
> I'd like for whatever the max number of pages to be displayed be the same in all of these situations. That way, the width of the bar doesn't change wildly as you progress across pages.

Okay. I will work on rendering same number of blocks for all cases.

> You covered a lot of edge cases really well, and it's a very readable component - so great job there.

Thanks!

components/Pagination/Pagination.js
kylemh: Right. I figured that, but was thinking about ways to simplify that component. I think we can just leave this alone though.

kylemh:
Lastly, how will these items be links? I think we can safely assume they'll only ever be internal links, using `next/link`, but just be aware of that!

prmsrswt:
I think we can pass an array of links and `currentPage` to `Pagination` and drop the `totalPages` as it will be equal to `pagesArray.length`.

The component will then handle the link assignment. Also by this way we can also remove `isClickable` from `PaginationItem` as only those component would be clickable which gets a link passed to it.

prmsrswt:
> I'd like for whatever the max number of pages to be displayed be the same in all of these situations. That way, the width of the bar doesn't change wildly as you progress across pages.

This should be fixed now. The default length is now `11` if the `totalPages` exceeds `11`, otherwise it would be equal to `totalPages`. Also it is configurable via an optional prop called `maxElements`.

> I would love a dev-only error that highlights when current page is > total pages

How do I go about displaying the error? I mean should I console log or return something like `<p>Error: invalid value</p>`

components/Pagination/Pagination.js
```diff
@@ -10,19 +10,40 @@ Pagination.propTypes = {
...
+  maxElements: 11,
 };
 
 // eslint-disable-next-line react/prop-types
-const PaginationItems = ({ currentPage, totalPages }) => {
-  const shouldTruncateStart = currentPage - 1 > 5;
-  const shouldTruncateEnd = totalPages - currentPage > 5;
+const PaginationItems = ({ currentPage, totalPages, maxLength }) => {
+  // maxElements is maximum length of the Pagination Bar, should be an odd integer, delault is 11
+  const maxElements = maxLength % 2 === 0 ? maxLength - 1 : maxLength;
```
kylemh: Lets drop `maxElements` since we don't need that aspect to be customizeable.

components/Pagination/Pagination.js
```diff
@@ -10,19 +10,40 @@ Pagination.propTypes = {
...
+  const elementsOnOneSide = (maxElements - 1) / 2;
```
kylemh: Then we can just hard-code this as a constant

`const MAX_CLICKABLE_VISIBLE = 7`

components/Pagination/Pagination.js
```diff
@@ -10,19 +10,40 @@ Pagination.propTypes = {
...
 
-  const paginationStart = shouldTruncateStart ? currentPage - 4 : 1;
-  const paginationEnd = shouldTruncateEnd ? currentPage + 4 : totalPages;
+  const shouldTruncateStart = currentPage - 1 > elementsOnOneSide && totalPages >= maxElements;
+  const shouldTruncateEnd =
+    totalPages - currentPage > elementsOnOneSide && totalPages >= maxElements;
+
+  const truncateStartOnly = shouldTruncateStart && !shouldTruncateEnd;
```
kylemh: can you booleanify these names as you did on 27 and 28?

components/Pagination/__tests__/Pagination.test.js
```diff
@@ -8,29 +8,39 @@ describe('Pagination', () => {
   afterEach(cleanup);
 
   it('should render with required props', () => {
-    createSnapshotTest(<Pagination currentPage={3} totalPages={5} />);
+    createSnapshotTest(<Pagination currentPage={3} totalPages={5} maxElements={11} />);
   });
 
-  it('should not have seperatorStart if currentPage - 1 is less than 6', () => {
-    const { queryByTestId } = render(<Pagination currentPage={6} totalPages={11} />);
+  it('should not have seperatorStart if currentPage - 1 is less than (maxElements + 1) / 2', () => {
```
kylemh: Any way you can make these test names a bit less mathy / bit more quickly skimmable?

prmsrswt: Yeah sure, also as we are going to drop `maxElements` the test names would be simplified.

components/Pagination/Pagination.js
prmsrswt: Okay :+1:

components/Pagination/Pagination.js
prmsrswt: Yeah it depends on `maxElements` and as we are dropping it, we can hard code this.

components/Pagination/Pagination.js
prmsrswt: Sure

kylemh:
@prmsrswt regarding the error, it would be dev only.

So just throw an error but only throw it when `process.env.NODE_ENV === 'development'`

components/Pagination/Pagination.js
```diff
@@ -0,0 +1,127 @@
...
+  totalPages: number.isRequired,
+};
+
+Pagination.defaultProps = {
+  className: undefined,
+};
+
+const getPagination = (currentPage, totalPages) => {
+  // maximum length of the Pagination Bar, should be an odd integer, delault is 11
+  const MAX_VISIBLE_ELEMENTS = 11;
```
kylemh: Didn't mean to make you create a new variable! I was tryna get you to drop `ELEMENTS_ON_ONE_SIDE` and use `const MAX_VISIBLE_ELEMENTS = 7`

components/Pagination/Pagination.js
```diff
@@ -0,0 +1,127 @@
...
+  return {
+    paginationStart,
+    paginationLength,
+    shouldTruncateStart,
+    shouldTruncateEnd,
+  };
+};
+
+// eslint-disable-next-line react/prop-types
+const PaginationItems = ({ currentPage, totalPages }) => {
```
kylemh: You can drop the lint disable and delete `className` as a possible prop!

components/Pagination/Pagination.js
prmsrswt: I am actually not sure how to go about it. Are we trying to decrease the length of Pagination from `11` to `7` and drop an unnecessary variable?

prmsrswt: Something like
```javascript
const MAX_VISIBLE_ELEMENTS = 7;

  const isTruncatingRequired = totalPages >= MAX_VISIBLE_ELEMENTS;

  const isLeftSideLengthy = currentPage - 1 > (MAX_VISIBLE_ELEMENTS - 1) / 2;
  const isRightSideLengthy = totalPages - currentPage > (MAX_VISIBLE_ELEMENTS - 1) / 2;

  const shouldTruncateStart = isTruncatingRequired && isLeftSideLengthy;
  const shouldTruncateEnd = isTruncatingRequired && isRightSideLengthy;

  const truncateStartOnly = shouldTruncateStart && !shouldTruncateEnd;
  const truncateEndOnly = !shouldTruncateStart && shouldTruncateEnd;

  let paginationStart;
  let paginationEnd;

  if (truncateStartOnly) {
    paginationStart = totalPages - MAX_VISIBLE_ELEMENTS + 3;
    paginationEnd = totalPages;
  } else if (truncateEndOnly) {
    paginationStart = 1;
    paginationEnd = MAX_VISIBLE_ELEMENTS - 2;
  } else {
    paginationStart = shouldTruncateStart ? currentPage - (MAX_VISIBLE_ELEMENTS - 5) / 2 : 1;
    paginationEnd = shouldTruncateEnd ? currentPage + (MAX_VISIBLE_ELEMENTS - 5) / 2 : totalPages;
  }
```

kylemh: Ah whoops. I didn't realize `MAX_VISIBLE_ELEMENTS` was being used outside of defining `ELEMENTS_ON_ONE_SIDE`

components/Pagination/Pagination.js
kylemh: So `shouldOnlyTruncateStart` and `shouldOnlyTruncateEnd`

components/Pagination/Pagination.js
prmsrswt: Yeah actually both variables are used as there are lot of edge cases to take care of.
> Nothing to do here then!

Okay cool.

components/Pagination/Pagination.js
```diff
@@ -0,0 +1,127 @@
...
+  );
+};
+
+function Pagination({ className, currentPage, totalPages }) {
+  const isCurrentPageTooSmall = currentPage < 1;
+  const isCurrentPageTooBig = currentPage > totalPages;
+
+  const isCurrentPageInvalid = isCurrentPageTooSmall || isCurrentPageTooBig;
+
+  const isDevelopmentEnvironment = process.env.NODE_ENV !== 'production';
```
kylemh: This isn't necessarily true, because it could be a custom `NODE_ENV` or `NODE_ENV` could be `'test'`. I think you can just inline this one on not worry about making a variable for it!

components/Pagination/Pagination.js
prmsrswt: I guess if we drop the lint disable it would require us to define prop types for currentPage and totalPages, but they are already verified when we require them in `Pagination` component.

components/Pagination/Pagination.js
```diff
@@ -0,0 +1,127 @@
...
+
+  if (isDevelopmentEnvironment && isCurrentPageInvalid) {
+    throw new Error('Invalid value for prop currentPage');
```
kylemh: You could use the variables you just defined to make this error super helpful and to short-circuit evaluation

for instance:
```javascript
// Define some error message object outside of the method an export it so you can use it in unit tests
const devErrors = {
  currentPageValue: value => `The value passed for currentPage is ${value}.`,
  currentPageTooSmall: '"currentPage" cannot be less than 1.',
  currentPageTooBig: '"currentPage" cannot be larger than "totalPages"',
};

// back in Pagination definition
const isCurrentPageTooSmall = currentPage < 1;
if (isCurrentPageTooSmall) {
  const errorMessage = `${devErrors.currentPageValue(currentPage)} ${devErrors.currentPageTooSmall}`
  throw new Error(errorMessage);
}

const isCurrentPageTooBig = currentPage > totalPages;
if (isCurrentPageTooBig) {
  const errorMessage = `${devErrors.currentPageValue(currentPage)} ${devErrors.isCurrentPageTooBig}`
  throw new Error(errorMessage);
}
```

components/Pagination/Pagination.js
kylemh: Ah it's cuz two JSX methods in one file. Okay, no worries!

components/Pagination/Pagination.js
prmsrswt: This is great! will modify the code according to this.

prmsrswt:
> Just need to add some tests that assert that an error is thrown!

Added!

> One annoying fun-fact is that when you're testing errors being thrown, you need to put your method into a callback within the expect to get the error. Check out existing error assertions for more on that.

This came handy. Wrote two tests for both situations where we expect Error thrown

components/Pagination/Pagination.js
prmsrswt: These names should be correct now.

components/Pagination/Pagination.js
```diff
@@ -0,0 +1,132 @@
...
+import RightAngleIcon from 'static/images/icons/FontAwesome/angle-right-solid.svg';
+import PaginationItem from './PaginationItem/PaginationItem';
+import styles from './Pagination.css';
+
+Pagination.propTypes = {
+  currentPage: number.isRequired,
+  totalPages: number.isRequired,
+};
+
+const developmentErrors = {
```
kylemh: If you export this, you can import it into the test file and reference it when making assertions about the errors. That way, you don't need to hard-code strings multiple times!

components/Pagination/Pagination.js
```diff
@@ -0,0 +1,132 @@
...
+          <PaginationItem key={totalPages} value={`${totalPages}`} testId={`${totalPages}`} />
+        </>
+      )}
+    </>
+  );
+};
+
+function Pagination({ currentPage, totalPages }) {
+  const isCurrentPageTooSmall = currentPage < 1;
+  if (isCurrentPageTooSmall && process.env.NODE_ENV !== 'production') {
```
kylemh: Instead of checking `process.env.NODE_ENV !== 'production'` for both, you can just wrap all of the dev error stuff into one conditional:

```javascript
if (process.env.NODE_ENV !== 'production') {
  const isCurrentPageTooSmall = currentPage < 1;
  if (isCurrentPageTooSmall) {
  // etc...

}
```

components/Pagination/Pagination.js
prmsrswt: Nice. I have pushed with these changes.

components/Pagination/Pagination.js
prmsrswt: Done :+1:

PR #868, opened 2019-10-10
Converted all forms to single column layout

### LINKED ISSUE #854: Make the form flow in a single direction.
# Bug Report
Currently the SignUp form is designed like this

image

The form should flow in one direction from top to bottom.

## Why is it bad?

It is a good UX practise to design the form with single columns instead of multi-column

<blockquote>
5. Avoid placing questions side-by-side.

Eye-tracking studies have shown that simple one-column layouts are better than multi-column layouts with questions positioned side-by-side.

The only exception to this rule is when asking for dates (day, month, year) or time (hours and minutes), where multiple fields are expected to be on one line.
</blockquote>

Source- https://www.ventureharbour.com/form-design-best-practices/

Although there is a blog saying that multi-column form worked great for them but they were having 13 fields so it was very long.

Since the registration form just has 7 fields. It would be better to make them single column.

Also it would not be longer than this since the required error will go to the right side

image

itaditya:
# Description of changes

With this PR all forms will get converted into single form layout which is a better UX.

# Issue Resolved

Fixes #854 

## Screenshots/GIFs

operation-code-registration-form

### DIFF
--- a/components/ChangePasswordForm/ChangePasswordForm.css
+++ b/components/ChangePasswordForm/ChangePasswordForm.css
@@ -4,8 +4,8 @@
 
 .row {
   display: flex;
-  flex-wrap: wrap;
-  justify-content: center;
+  flex-direction: column;
+  align-items: center;
 }
 
 .topMargin {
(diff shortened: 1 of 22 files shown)


kylemh:

Built with commit 2611b009da1291c071749186b4f4dfdb4228e4aa

https://deploy-preview-868--operation-code-storybook.netlify.com

itaditya:
Thanks @kylemh.

PR #881, opened 2019-10-14
Custom ESLint Rule

### LINKED ISSUE #867: Use existing or create custom lint rule for proptype definitions to exist above function declaration
# Feature request
I **don't** want to see:

```jsx
const SomeComponent = (props) => <div {...props} />

SomeComponent.propTypes = {}
```

I want to see:

```jsx
SomeComponent.propTypes = {}

function SomeComponent(props) {
  return (
    <div {...props} />
  );
}
```

## Proposed solution
Create a custom lint rule to automate my above preferences.

I couldn't find one, but if you can find an existing lint rule to do this, please do that!

chynh:
# Description of changes

Created custom eslint rule for ensuring proptype definitions exist above component declaration.
# Issue Resolved

Fixes #867 

## Screenshots/GIFs

Oct-14-2019 10-01-28


kylemh:

Built with commit 8dadfc46b39199c1ca9e87456f2342084dbd0474

https://deploy-preview-881--operation-code-storybook.netlify.com

eslint/index.js
```diff
@@ -0,0 +1,57 @@
...
+              return false;
+            };
+            // PropType node
+            const propertyTypeNode = node.left.object;
+            // start loc of PropType node
+            const compareValue = propertyTypeNode.start;
+            if (node.left.property.name === 'propTypes') {
+              // find the component the proptype declaration was intended for
+              // Had to implement it this way otherwise eslint would error
+              const targetNode =
```
kylemh: This is a good use-case for `lodash.get`!

```js
import get from 'lodash/get';

// ...

const body = get(node, 'parent.parent.body', []);
const targetNode = body.find(element => findDeclaration(element));
```

eslint/index.js
```diff
@@ -0,0 +1,57 @@
...
+                  (singleNode.declarations &&
+                    singleNode.declarations[0].id.name === propertyTypeNode.name)
+                ) {
+                  return true;
+                }
+              }
+              return false;
+            };
+            // PropType node
+            const propertyTypeNode = node.left.object;
```
kylemh: You can disable the lint rule for "no abbreviations" here becuz it's an API we don't control.

```suggestion
            // delete line 32 comment
            const propTypeNode = node.left.object; // eslint-disable unicorn/prevent-abbreviations
```

eslint/index.js
```diff
@@ -0,0 +1,57 @@
+module.exports = {
+  rules: {
+    'proptype-definition': {
+      create(context) {
+        return {
+          AssignmentExpression(node) {
+            // helper func which targets possible react components
+            const findDeclaration = singleNode => {
```
kylemh: Can you extract this to the top of the file? No need to add the comment -maybe just change the name to indicate that it always returns a boolean! Something like: `hasDeclaration`

Either way, it will just read cleaner if it's separated from the actual rule.

eslint/index.js
```diff
@@ -0,0 +1,57 @@
+module.exports = {
+  rules: {
+    'proptype-definition': {
```
kylemh: Let's call this `proptype-definition-above-fn`

.eslintrc.js
```diff
@@ -84,6 +84,7 @@ module.exports = {
         svg: 'always',
       },
     ],
+    'custom-rules/proptype-definition': 2,
```
kylemh: can you do `'error'` instead of `2`

eslint/index.js
```diff
@@ -0,0 +1,57 @@
...
+                  // deal with variable declaration or function statement
+                } else if (
+                  (singleNode.id && singleNode.id.name === propertyTypeNode.name) ||
+                  (singleNode.declarations &&
+                    singleNode.declarations[0].id.name === propertyTypeNode.name)
+                ) {
+                  return true;
+                }
+              }
+              return false;
```
kylemh: This is a good use-case for `lodash.get`!

```js
if (!singleNode.type.match(/[Function|Variable|Class|Export(Named|Default)]Declaration/)) {
  return false;
}

const isDefaultOrNamedExport =
  get(singleNode, 'declaration.id.name') === propertyTypeNode.name ||
  get(singleNode, 'declaration.id.name.declarations[0].id.name') ===
    propertyTypeNode.name;

if (isDefaultOrNamedExport) return true;

const isVariableOrFunctionStatement =
  get(singleNode, 'id.name') === propertyTypeNode.name ||
  get(singleNode, 'declarations[0].id.name') === propertyTypeNode.name;

if (isVariableOrFunctionStatement) return true;
```

package.json
```diff
@@ -143,6 +143,7 @@
     "eslint-plugin-prettier": "^3.1.1",
     "eslint-plugin-react": "^7.16.0",
     "eslint-plugin-unicorn": "^12.1.0",
+    "eslint-plugin-custom-rules": "file:eslint",
```
kylemh: How does this work?

chynh:
Hey @kylemh, yeah I was pretty stoked to get this working. I did not read any documentation although I really should have lol. I read some Medium articles on creating custom rules and just figured it out by playing around. 

Here are the articles that I referenced:
- [Custom ESLint Rules For Faster Refactoring](https://medium.com/neighborhoods-com-engineering-blog/custom-eslint-rules-for-faster-refactoring-2095e69bde08)
- [Create custom ESLint rules in 2 minutes](https://blog.webiny.com/create-custom-eslint-rules-in-2-minutes-e3d41cb6a9a0)

package.json
chynh: This is basically telling where to look for the custom ESLint plugin which is in "eslint" folder that I added. So when running yarn, the plugin will be copied into node_modules folder. More detailed explanation can be found in the first article that I referenced.

eslint/index.js
chynh: okay, I will make the change later today

.eslintrc.js
chynh: Sure.

eslint/index.js
chynh: Sure.

eslint/index.js
chynh: Will do.

eslint/index.js
chynh: Got it.

package.json
kylemh: That's very cool. I'm only really worried about how this works in CI and/or on Windows machines.

chynh: Ah okay. I suppose I could publish this as a npm package. What do you think?

kylemh: Let's see how it works for us!

kylemh:
I have a request that you try not to force push commits. I can't tell if history gets overridden and makes PRs a bit more difficult!

kylemh:
Got one last curve ball for you! What about `SomeFunction.defaultProps`?

chynh:
> I have a request that you try not to force push commits. I can't tell if history gets overridden and makes PRs a bit more difficult!

Got it. 

> Got one last curve ball for you! What about `SomeFunction.defaultProps`?

Yeah. I thought this might come up. I don't think it would be too difficult implementing this as the logic would be pretty much the same. I'll make some changes to the rule so we got defaultProps covered as well.

kylemh:
It's failing in CI because the rule can't be found.

kylemh:
here's what i see

Screen Shot 2019-10-17 at 11 57 55 AM

chynh:
Hmm...I think this is exactly what you were concerned about earlier.

> That's very cool. I'm only really worried about how this works in CI and/or on Windows machines.

I guess the rule is not being installed during ci, causing this issue. 

I have looked into this [article](https://docs.npmjs.com/files/package.json#local-paths) to decide what would be the best approach is here. I can try adjusting the local path of the rule or alternatively put the file up on Github and list the Git URLs as Dependency. What do you think?

kylemh:
Personally I think the easiest thing would be to make an NPM lib for these new custom rules.

Do you want to publish Operation Code's first NPM library?

chynh:
Just checking, are we ready to publish the rule or is there something else that needs to be done?

kylemh:
You should now be able to install this via: `yarn add -D @operation_code/eslint`

chynh:
I had to make a few modifications since the rule was not working. Also, I added a short readme in the repo as well. Could you go ahead and re-publish it?

kylemh:
https://www.npmjs.com/package/@operation_code/eslint-plugin-custom-rules 👍 

I changed the import path for lodash in that project btw!

kylemh:
Great work! Pulled it down locally and it worked like a charm.

PR #930, opened 2019-11-06
Migrate to hooks, need help with `nextContextMock`

### LINKED ISSUE #845: Migrate to Hooks
The code presently uses class components & lifecycle methods.
However, React 16.8.0 release introduces a more intuitive & functional approach towards development which is more likely to be used in the future.

## Steps to resolve the issue:
Replace Class and Class-based methods (state, lifecycle, PureComponent etc.) to functional approach (useState, useEffect, memo, etc.)

## Keep in mind:
https://github.com/OperationCode/front-end/issues/867

## Note

Comment the names of the 5 components you plan on migrating before tackling them so you don't step on anybodys' toes.

rusye:
# Description of changes
Transition to React Hooks

# Issue Resolved
Fixes #845

# Help
I'm not 100% positive that my solution for `nextContextMock` will work properly


kylemh:

Built with commit 351fd55df40e793a451028a1bd25b5a81eade510

https://deploy-preview-930--operation-code-storybook.netlify.com

rusye:
Anything related to my code that I have to fix to make this work?

decorators/withAuthSync/withAuthSync.js
```diff
@@ -1,40 +1,35 @@
-import React from 'react';
+import { useEffect } from 'react';
 import Router from 'next/router';
 import { authenticate } from 'common/utils/auth-utils';
 import getDisplayName from 'decorators/getDisplayName';
 
-const withAuthSync = WrappedComponent =>
-  class extends React.Component {
-    static displayName = `withAuthSync(${getDisplayName(WrappedComponent)})`;
+const syncLogout = event => {
```
kylemh: Can you define this within the hook?

decorators/withAuthSync/withAuthSync.js
```diff
@@ -1,40 +1,35 @@
...
-    }
-
-    syncLogout = event => {
-      if (event.key === 'logout') {
-        Router.push('/login');
-      }
-    };
+  // eslint-disable-next-line unicorn/prevent-abbreviations
+  const componentProps =
+    useWithAuthSync.getInitialProps && (await useWithAuthSync.getInitialProps(ctx));
```
kylemh: You'll wanna think about this differently if you're tryna create a hook. It looks like you're trying to do a decorator AND a hook.

`useWithAuthSync.getInitialProps` isn't the same thing as `WrappedComponent.getInitialProps`

Let's just consider `useAuthSynchronize` a method that gets called at the top of the definition of every page fn that used to be wrapped with the decorator.

rusye: I have no idea what I was doing that day... Going to work on it today.

kylemh: You were giving it a good go!

rusye: I've looked at one too many articles and I'm still lost, this is the best I could do
```function useWithAuthSync() {
  useEffect(() => {
    displayName(() => `withAuthSync(${getDisplayName(useWithAuthSync)})`);

    syncLogout(event => {
      if (event.key === 'logout') {
        Router.push('/login');
      }
    });

    getInitialProps(async ctx => {
      const token = authenticate(ctx);

      // eslint-disable-next-line unicorn/prevent-abbreviations
      const componentProps =
        useWithAuthSync.getInitialProps && (await useWithAuthSync.getInitialProps(ctx));

      return { ...componentProps, token };
    });

    window.addEventListener('storage', syncLogout);

    return () => {
      window.removeEventListener('storage', syncLogout);
      window.localStorage.removeItem('logout');
    };
  }, []);
}

export default useWithAuthSync;

kylemh: For one, we won't need to worry about display name anymore. This is a hook that can get called within any component - it's not longer wrapping an existing component.

You won't want to define `getInitialProps` within `useEffect` you'll want to call the `getInitialProps` property that exists upon the component calling this hook.

Does that help?

Above all else, make sure what you're writing is working! You'll want to ensure that you can't visit `/profile` unless you've logged in - for example.

kylemh: We won't be able to make this a hook https://spectrum.chat/next-js/general/static-getinitialprops-in-react-hooks~bf6f053e-3248-410e-b0bc-f200d3092b86

It needs to remain a decorator.

PR #943, opened 2019-11-18
Integrate Cypress Retry Plugin 

### LINKED ISSUE #937: Integrate Cypress Retry plugin in CI
# Feature request
One way I've been resolving flaky CI is by simply manually re-running tests.

## Proposed solution
Put this: https://github.com/Bkucera/cypress-plugin-retries into our repo and enable it only for CI if possible and retry stuff 3 times.

chynh:
# Description of changes
- Installed ``cypress-plugin-retries`` and configured it so that a flaky test would re-run if it fails
- Created ``retry.spec.js`` to ensure tests are re-running up to three times.
- Modified  ``jest.config.js`` and ``nyc.config.js`` so that ``components/Timeline/historyData.js`` and ``components/UpdateProfileForm/steps/index.js
`` are properly ignored.

# Issue Resolved

Fixes #937


kylemh:

Built with commit a7aa5d177e452aef6921ad2fd3fe63bbaa341ac5

https://deploy-preview-943--operation-code-storybook.netlify.com

jest.config.js
```diff
@@ -44,6 +44,7 @@ module.exports = {
     '<rootDir>/common/utils/api-utils.js',
     '<rootDir>/components/ZipRecruiterJobs/ZipRecruiterJobs.js',
     '<rootDir>/components/Press/PressLinks/Articles.js',
+    '<rootDir>/components/Timeline/historyData.js',
```
kylemh: good catch!

nyc.config.js
```diff
@@ -6,5 +6,10 @@ const removeRootDirectory = pathArray => {
 
 module.exports = {
   include: [...removeRootDirectory(config.collectCoverageFrom), 'pages/**/*.js'],
-  exclude: [...removeRootDirectory(config.coveragePathIgnorePatterns), 'pages/api/__coverage__.js'],
+  exclude: [
+    ...removeRootDirectory(config.coveragePathIgnorePatterns),
+    'pages/api/__coverage__.js',
+    'common/**/index.js',
+    'components/**/index.js',
```
kylemh: Why not use `config.collectCoverageFrom` in some way?

cypress/integration/retry.spec.js
```diff
@@ -0,0 +1,8 @@
+describe('Retrying Cypress tests', () => {
```
kylemh: I changed the verbiage in this test to be crystal clear.

nyc.config.js
chynh: Sure.

I noticed that just removing ``<rootDir>`` was not working in this case since ``collectCoverageFrom`` is an array of regexp pattern strings whereas ``exclude`` is an array of glob patterns. 

I suppose I can use a function to turn strings form ``collectCoverageFrom`` into glob patterns.

kylemh: That's ideal. Definitely want to get away from defining these strings twice.

nyc.config.js
```diff
@@ -1,7 +1,7 @@
 const config = require('./jest.config');
 
 const removeRootDirectory = pathArray => {
-  return pathArray.map(path => path.slice(10));
+  return pathArray.map(path => path.replace('(.*)', '**').slice(10));
```
kylemh: Did this do anything? We still want to read `collectCoverageFrom`, right?

chynh: Oh my bad. I totally misunderstood the part about using ``collectCoverageFrom``.

I believe with this modification, ``components/UpdateProfileForm/steps/index.js`` is properly ignored during Cypress testing.

What I am really targeting with this modification are ``'<rootDir>/common/(.*)/index.js'`` and ``'<rootDir>/components/(.*)/index.js'`` from ``coveragePathIgnorePatterns`` and basically replacing ``(.*)`` with ``**``. Any other paths in ``collectCoverageFrom`` and  ``coveragePathIgnorePatterns`` will not be affected since those two are the only ones with that pattern.

But we can definitely use ``collectCoverageFrom`` as we initially discussed. Let me know what you think!

kylemh: I see now. I guess I'm just surprised we didn't do:

```js
const removeRootDirFromPaths = (pathsArray) => {
  return pathsArray.map(path => path.replace('<rootDir>/', ''));
}

// ...

  include: ['pages/**/*.js'],
  exclude: [
    ...removeRootDirFromPaths(config.collectCoverageFrom), 
    ...removeRootDirFromPaths(config.coveragePathIgnorePatterns)),
  ]
```

kylemh: Or... do we even need the `exclude` array?

chynh: Correct. I believe we won't need ``exclude`` array if we only included ``pages/**/*.js``

But, I think we need to at least exclude ``pages/api/__coverage.js__`` there.

And I believe the overall test coverage result will benefit if we include the ``config.collectCoverageFrom``, at least I am certain that including ``common/**/*.js`` and ``decorators/**/*.js`` has increased the total percentage. 

So, I think it really comes down to what we include in the coverage and depending on that, we can decide what to exclude from the result.

kylemh: Wouldn't the `jest` config capture that though, and then the reports get merged.

chynh: Yep. Jest definitely does.

But I am pretty sure that test coverage for certain files such as ``decorators/withAuthSync/withAuthSync.js`` went up after the merge when those were included in the Cypress test. 

To be sure, I think we would be able to check if there is any difference in the result on CodeClimate if I make a new commit with the modified config as below.

```
include: ['pages/**/*.js'],
exclude: ['pages/api/__coverage__.js']
```

kylemh: Yes, coverage goes up because everything gets called again. In reality, `withAuthSync` can have it's own unit/integration test with Jest, but it doesn't have one. Although, it would be hard to test without the context of full-blown routing...

Alright, let's go with this - final answer:

```
include: ['pages/**/*.js', 'decorators/withAuthSync/withAuthSync.js'],
exclude: ['pages/api/__coverage__.js']
```

kylemh:
OKAY! I see what I want now, based off of the diff.

```
include: ['pages/**/*.js', ... removeRootDirFromPaths(config.collectCoverageFrom)],
exclude: ['pages/api/__coverage__.js']
```

chynh:
> OKAY! I see what I want now, based off of the diff.
> 
> ```
> include: ['pages/**/*.js', ...removeRootDirFromPaths(config.collectCoverageFrom)],
> exclude: ['pages/api/__coverage__.js']
> ```
> 
> thanks for your patience 😂

No worries!

PR #1025, opened 2020-01-20
Add minimum node version

aaron-suarez:
# Description of changes
This PR updates the contributing guide with the minimum node version required.

# Issue Resolved
I normally use docker so I don't keep my native install of node up to date (also, it's been a while since I've done JS development, so it probably would have been outdated anyway). When I ran `yarn` to install dependencies, I got this error:
```bash
error @svgr/webpack@5.0.1: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.10.0"
error Found incompatible module.
```
This let's people know when they need to upgrade in case they have old versions of node that they're working with.

## Screenshots/GIFs

N/A

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -195,8 +195,8 @@ VERY IMPORTANT: Required versions of tools used within the repo are described [h
 _You can check to see your versions like so:_
 
 - Run `node -v`. You can download the latest LTS release of node at [nodejs.org](https://nodejs.org) or you can use [nvm](https://github.com/creationix/nvm) to be able to switch between node versions easily for many projects. If you use Windows, you will need to use [nvm-windows](https://github.com/coreybutler/nvm-windows) instead.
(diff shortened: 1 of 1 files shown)


kylemh:
The text you added is arbitrary and will require updating every time we change major node versions. I’m really opposed to adding documentation that needs to stay in sync with code unless we have no alternative. The comment about the LTS should’ve told you to use at least 10, since 8 is no longer supported.

Is it okay if I close this, or do you feel strongly that the line should remain?

aaron-suarez:
I don't feel that strongly about it, but I do think that we should have consistency. Can we remove the minimum npm and yarn versions then, if we don't want to have to keep this document in sync?

kylemh:
Sure! Wanna make those changes here?

aaron-suarez:
Yup! I'm mobile right now but I should be back by my computer in about an hour or two. I'll get those changes made and then if you're available we can do a slack call about the /resources page

PR #1100, opened 2020-04-26
Resources page search

manthonyg:
# Description of changes
Creating the search for the resources page.


kylemh:

Built with commit 4d43b875e0cd3f83095b64281122a79b09214ed3

https://deploy-preview-1100--operation-code-storybook.netlify.app

manthonyg:
I wanted to get some stuff done on this and maybe get it to that illusive MVP spot so people can use it. Problems that still need to be addressed:

1. The URL in the address bar is always one "step" behind. This can quickly get confusing for someone that switches from searching for free course to a custom query or something. I was under the impression that `router.push` should immediately update the browser history/the path in the url bar but maybe not?

2. Although everything _looks_ good to me when using the multiselect (multiple languages like `&languages=javascript&languages=python`, it will only work correctly for one language. Haven't found a workaround.

3. As far as handling the case where `fetchedResources or fetchedPages === 0` and being able to show a 'no resources found' alert, the server sends back a 404 if either of those cases are true and there is no response data to work off of, just 'Failed to load response data'. Don't know if I am missing something elementary  here, I talked to aaron and he is taking a look into it.

pages/resources/[page].js
```diff
@@ -1,57 +1,170 @@
...
+      return getResourcesPromise({ page, category, languages, paid });
+    }
+    return getResourcesBySearch({ page, q, paid });
+  };
+
+  useEffect(() => {
+    setErrorMessage(null);
+
+    // eslint-disable-next-line no-restricted-globals
+    if (isNaN(currentPage)) {
```
Comment (kylemh) [inline pages/resources/[page].js:94]: I think we can just force a redirect to page 1. I doubt a user would care to see this error message and fix out the best thing to do.

Comment (manthonyg) [reply pages/resources/[page].js:94]: Ok, yeah makes sense- so you want that to be server side thing in `now.json` using regex to determine if they are requesting an invalid page followed by redirect, or just push to `/resources/1` on client?

Comment (kylemh) [reply pages/resources/[page].js:94]: Just push 👍

Comment (manthonyg) [reply pages/resources/[page].js]: So this was a bit tricky for me since whether I try to use `isNumber, isNan`, or any other relevant type checker I can think of, even on resources/1 it will flash `NaN` at least once, long enough stick me in an infinite loop. I can't use router outside of `useEffect` to redirect, and the dependency is the query. What do?

manthonyg:
update:
**little victories**
1. 404s are good now and show an error for no resources
2. url bar stays in sync with the requests
3. paid param will spread itself into existing parameters vs being a stand alone
4. did my best to make when the selects reset themselves make sense.

**big L's**
- [ ] -   I cannot figure out how to control the state on the input form, probably because I don't understand Formik at all.  Right now it just resets the form on submit. Probably not great for UX.
- [ ] -  `Suspense` doesn't seem like a good option for SSR
- [ ] - as commented above, the redirect on client for `resources/somethingThatIsNaN` is problematic because it sticks me in an infinite loop. I cant use `router` outside of `useEffect` and `useEffect` takes the query as a dep. It always flashes `currentPage` as `NaN` at least on first render.

pages/resources/[page].js
Comment (kylemh) [reply pages/resources/[page].js]: That makes sense cuz we're not returning, so the rest of the UI attempts to render a tick before the push occurs.

```
if (isNaN(currentPage)) {
  return router.push()
}
```

let's slap this IMMEDIATELY after the `page` is parsed as an integer, so that nothing else evaluates.

kylemh:
for future reference, no need to make it a draft PR :)

common/utils/api-utils.js
```diff
@@ -3,6 +3,8 @@ import lodashGet from 'lodash/get';
 import { networkErrorMessages } from 'common/constants/messages';
 import { apiUrl } from 'common/config/environment';
 import { setAuthorizationHeader } from 'common/utils/cookie-utils';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import qs from 'qs';
```
kylemh, line 7: if you run `yarn add qs`, you can remove the lint statement

common/utils/api-utils.js
```diff
@@ -50,10 +53,15 @@ export const get = async (path, { token, parameters } = {}, axiosClient = Operat
       headers: setAuthorizationHeader(token),
       cancelToken: abort.token,
       params: parameters,
+      paramsSerializer: parameters_ => qs.stringify(parameters_, { arrayFormat: 'repeat' }),
```
kylemh, line 56: Can you leave a comment about why this is here 🙏

common/utils/api-utils.js
```diff
@@ -50,10 +53,15 @@ export const get = async (path, { token, parameters } = {}, axiosClient = Operat
...
     })
     .then(response => {
       clearTimeout(connectionTimeout);
       return response;
+    })
+    .catch(error => {
+      clearTimeout(connectionTimeout);
+      return error.response.data;
```
kylemh, line 64: Can you just return the error? As it stands, we've so far been good about handling every exception on a case-by-case basis.

components/Cards/ResourceCard/__tests__/ResourceSkeletonCard.test.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { SKELETON_CARD } from 'common/constants/testIDs';
+import ResourceSkeletonCard from '../ResourceSkeletonCard';
+
+describe('ResourceSkeletonCard', () => {
+  it('renders the correct amount of skeletons', () => {
+    const component = render(<ResourceSkeletonCard numberOfSkeletons={5} />);
+    const Skeletons = component.queryAllByTestId(SKELETON_CARD);
+    expect(Skeletons).toHaveLength(5);
```
kylemh, line 10: the `5` here should be tied to a variable to point out that they're the same

pages/resources/[page].js
```diff
@@ -1,75 +1,254 @@
...
+      actions.setSubmitting(false);
+    }, 500);
+  };
+
   useEffect(() => {
-    getResourcesPromise({ page: query.page })
+    Promise.all([getResourcesByCategories(), getResourcesByLanguages()])
+      .then(([categoriesResponse, languagesResponse]) => {
+        const {
+          data: { data: categoriesData },
```
Comment (kylemh) [inline pages/resources/[page].js:106]: It might be useful to bring up the weirdness of `data.data` to Aaron. If we keep on adding more stuff it's only gonna get more obtuse.

pages/resources/[page].js
```diff
@@ -1,75 +1,254 @@
...
+              label: categoryObject.name,
+            };
+          }),
+        );
+      })
+      .catch(error => {
+        setErrorMessage(handleError(error));
+      });
+
+    setErrorMessage(null);
```
Comment (kylemh) [inline pages/resources/[page].js:133]: Wouldn't this clear any error messages even even if an asynchronous one was supposed to exist? I think you mean to call this during submit and on reset

pages/resources/[page].js
```diff
@@ -1,75 +1,254 @@
...
   const { pathname, query } = router;
-  const currentPage = parseInt(query.page, 10);
+  const { page, category, languages, paid, q } = query;
+  const currentPage = parseInt(page, 10);
 
-  if (!isNumber(currentPage)) {
-    // TODO: Handle situation where user tried some funny business, like `/resources/fuck-you`
-    throw new Error(`'currentPage' is not a number`);
+  // eslint-disable-next-line no-restricted-globals
+  if (page && isNaN(page)) {
```
Comment (kylemh) [inline pages/resources/[page].js:59]: You haven't imported `isNaN` from lodash which is why it's complaining about the global. Do so, and you can remove the lint disable.

pages/resources/[page].js
```diff
@@ -1,75 +1,254 @@
...
+  getResourcesByLanguages,
+  getResourcesBySearch,
+} from 'common/constants/api';
+import { Field, Formik } from 'formik';
+import Form from 'components/Form/Form';
+import Input from 'components/Form/Input/Input';
+import Button from '../../components/Button/Button';
 import styles from '../styles/resources.module.css';
+import Select from '../../components/Form/Select/Select';
+import Alert from '../../components/Alert/Alert';
```
Comment (kylemh) [inline pages/resources/[page].js:25]: Will you please use our absolute imports here?

kylemh:
Side note. It looks like we'll want the validation stylings on inputs to be prop-controlled.

As in... I don't think we need the borders to turn green in this form. I can make that a separate PR for ya

common/utils/api-utils.js
manthonyg: You got it 👍🏼

pages/resources/[page].js
Comment (manthonyg) [reply pages/resources/[page].js:133]: Totally. Thanks for catching it

kylemh:
Added: https://github.com/OperationCode/front-end/pull/1121

manthonyg:
Until I can figure out how to get cypress working again @aaron-suarez  recommended I manually run it through the ringer to see what comes up. I found two bugs:

1) anytime the pagination is truncated you can no longer click past the `...` without the query no longer persisting and you being directed to `resources/whateverPage#youClicked`. if it is not truncated and you visit the last page it is fine.

2) languages are working when requested with an array (like `language=python&language=javascript` would return resources with a language of python, javascript, or both) but categories will not work like this, the data is always whatever the first requested resource was

**Redirects**

Action - Navigate to `/resources`
Events- 
now.json redirects to `resources/1`
data displayed is first page containing 20 resources
URL- `resources/1`

Action - Navigate to `/resources/RANDOMSTRING`
Events- 
client redirects to `resources/1`
data displayed is first page containing 20 resources
URL- `resources/1`

**Using resources**

Action - Paginate to page `2`
Events - 
data displayed is second page containing 20 resources
URL- `resources/2`

Action - Paginate to last page (130)
Events -
data displayed is 130th page containing 5 resources
URL- `resources/130`

Action - Click on ‘back’ arrow
Events - 
data displayed is 129th page containing 20 resources
URL- `resources/129`

Action - Click on ‘forward’ arrow
Events -
data displayed is 130th page containing 5 resources
URL - `resources/130`

Action - Click ‘reset’ button from current search
Events - 
data displayed is 1st page containing 20 resources
URL- `resources/1`

**Using search**

Action - Type ‘javascript’ into search bar and click search or press enter
Events - 
data displayed is 1st page containing 20 resources matching ‘javascript’
URL - `resources/1?q=javascript`

Action - Click on ‘forward’ arrow
Events - 
data displayed is 2nd page containing 20 resources matching ‘javascript’
URL - `resources/2?q=javascript`

Action - Paginate to last page (8)
Events -
data displayed is 8th page containing 5 resources mataching ‘javascript'
URL- `resources/8?q=javascript`

Action - Click on ‘back’ arrow
Events -
data displayed is 7st page containing 20 resources matching ‘javascript’
URL - `resources/7?q=javascript`

Action - Click ‘reset’ button from current search
Events - 
search field clears
data displayed is 1st page containing 20 resources
URL- `resources/1`

**Using category**

Action - Select ‘Code Challenges’ in ‘Category’ select and click search
Events -
Category select components value is ‘Code Challenges’
data displayed is 1st page containing 20 resources with a category of ‘Code Challenges’
URL - `resources/1?category=code%20challenges`

Action - Click on ‘forward’ arrow
Events -
Category select components value is ‘Code Challenges’
data displayed is 2nd page containing 20 resources with a category of ‘Code Challenges’
URL - `resources/2?category=code%20challenges`

Action - Paginate to last page (3)
Events -
data displayed is 3rd page containing 5 resources with a category of ‘Code Challenges'
URL- `resources/8?category=code%20challenges`

Action - Click on ‘back' arrow
Events -
Category select components value is ‘Code Challenges’
data displayed is 2nd page containing 20 resources with a category of ‘Code Challenges’
URL - `resources/2?category=code%20challenges`

**** Action - Select ‘Books’ as a secondary category in ‘Category’ select and click search
Events -
Category select components value is ‘Code Challenges’
data displayed is 1st page containing 20 resources with a category of ‘Code Challenges’
URL - `resources/1?category=code%20challenges&category=books`

Action - Hit the ‘X’ on Code Challenges within the multi select and hit search
Events -
Category select components value is ‘Code Challenges’
data displayed is 1st page containing 20 resources with a category of ‘Books’
URL - `resources/1?category=books`

Action - Click ‘reset’ button from current search
Events -
Category select components value is the placeholder
data displayed is 1st page containing 20 resources
URL- `resources/1`


**Using languages**

Action - Select ‘Python' in ‘Languages’ select and click search
Events -
Category select components value is the placeholder
data displayed is 1st page containing 20 resources
URL- `resources/1?languages=python`

Action - Click on ‘forward' arrow
Events -
Language select components value is ‘Python'
data displayed is 2nd page containing 20 resources with a language of ‘Python'
URL- `resources/2?languages=python`

Action - Paginagte to last page (6)
Events -
data displayed is 6th page containing 4 resources with a language of ‘Python'
URL- `resources/6?languages=python`

Action - Click on ‘back’ arrow
Events -
data displayed is 5th page containing 20 resources with a language of ‘Python'
URL- `resources/5?languages=python`

Action - Click ‘reset’ button from current search
Events -
Language select components value is the placeholder
data displayed is 1st page containing 20 resources
URL- `resources/1`

**Using paid**

Action - Select ‘Paid' in ‘Cost’ select and click search
Events -
Paid select components value is the ‘Paid'
data displayed is 1st page containing 20 resources with a paid value of true
URL- `resources/1?paid=true`

Action - Click on ‘forward' arrow
Events -
data displayed is 2nd page containing 20 resources with a paid value of true
URL- `resources/2?paid=true`

Action - Paginagte to last page (15)
Events -
data displayed is 15th page containing 4 resources with a language of ‘Python'
URL- `resources/6?languages=python`

aaron-suarez:
Do you think we need some way to show what parameters are being used to display the current set of resources? I feel like if you're not paying attention, you might not realize that the list has been updated and that your search is complete. Maybe either we can slow it down and have like some sort of loading icon overlay so you see that your search is happening, or like have a label that says "Showing results for..."

I don't know if that's necessary but I just feel like it's hard keeping track of where I am

aaron-suarez:
I think we've been over this, but I don't like that the resource card doesn't show all of the data we have on a resource. Specifically, I would like to see the category and language(s) associated with the resource in addition to the notes. Remind me, why are we holding off on that?

aaron-suarez:
Ooh, so you know how the upvote and downvote thumbs buttons don't do anything? Should we have a little info bubble pop up letting the user know that this feature is coming soon? That way we can deploy this and start sharing it around but people aren't turned off by the fact that it doesn't work. Maybe a call to action to help out with the site, linking them to the github issue or the repo

components/Pagination/Pagination.js
```diff
@@ -90,7 +90,7 @@ const PaginationItems = ({ currentPage, pathname, query, totalPages }) => {
     <>
       {shouldTruncateStart && (
         <>
-          <PaginationItem key="1" value={1} testId="1" pathname={pathname}>
+          <PaginationItem key="1" value={1} query={query} testId="1" pathname={pathname}>
```
kylemh, line 93: Looks like we can remove changes to this file 👏

cypress/integration/resources.spec.js
```diff
@@ -0,0 +1,237 @@
+/* eslint-disable cypress/no-unnecessary-waiting */
```
kylemh, line 1: Can we remove this?

cypress/integration/resources.spec.js
```diff
@@ -0,0 +1,237 @@
...
+  const CategorySelect = 'input#react-select-category-input';
+
+  beforeEach(() => {
+    cy.visitAndWaitFor('/resources/1');
+    cy.get('h1').should('have.text', 'Resources');
+  });
+
+  it('redirects on /resources to resources/1', () => {
+    cy.visitAndWaitFor('/resources');
+    cy.location('pathname').should('eq', '/resources/1');
```
kylemh, line 21: can we also assert the URL has been changed?

cypress/integration/resources.spec.js
```diff
@@ -0,0 +1,237 @@
...
+  });
+
+  it('loads many resource cards on resources/1', () => {
```
kylemh, line 24:
```suggestion
  it('renders many resource cards', () => {
```

cypress/integration/resources.spec.js
```diff
@@ -0,0 +1,237 @@
...
+    cy.findAllByTestId(RESOURCE_CARD).should('have.length', 20);
+  });
+
+  it('renders the second page of results when using pagination to visit page 2', () => {
+    cy.findByTestId('2').click();
+    cy.location('pathname').should('eq', '/resources/2');
+  });
+
+  it('renders the next page of results when using pagination to visit next page', () => {
+    cy.findByTestId('rightAngle').click();
```
kylemh, line 35: I feel like we can improve the testID name and also add it to the file for testIDs. Something like "NEXT_PAGE_BUTTON"

cypress/integration/resources.spec.js
```diff
@@ -0,0 +1,237 @@
...
+    cy.location('pathname').should('eq', '/resources/2');
+  });
+
+  it('renders the previous page of results when using pagination to visit previous page', () => {
+    cy.visitAndWaitFor('/resources/2');
+    cy.findByTestId('leftAngle').click();
```
kylemh, line 41: same here

cypress/integration/resources.spec.js
```diff
@@ -0,0 +1,237 @@
...
+
+  it('renders the last page of results when paginating to the last page', () => {
+    let pageValue;
+    cy.get('ol')
+      .find('li')
+      .eq(-2)
+      .click()
+      .then($li => {
+        pageValue = $li.text().match(/[0-9]+/);
+      });
```
kylemh, line 53: Let's apply a test-id to the specific page number buttons

cypress/integration/resources.spec.js
```diff
@@ -0,0 +1,237 @@
...
+    cy.location().should(loc => {
+      expect(loc.pathname).to.eq(`/resources/${pageValue}`);
+      expect(loc.search).to.eq('');
+    });
+  });
+
+  it('will allow a user to search matching resources by input', () => {
+    cy.findByTestId(RESOURCE_SEARCH).click().type('javascript', { force: true }).type('{enter}');
+
+    cy.findByTestId('2').click();
```
kylemh, line 63: Hmmm what has a testid of 2?

cypress/integration/resources.spec.js
manthonyg: all set, made them `NEXT_PAGE_BUTTON` and `PREV_PAGE_BUTTON` respectively and put into `testIds` dir.

components/Pagination/Pagination.js
kylemh: Nvm. I didn't see that `query` was used.

cypress/integration/resources.spec.js
kylemh: 🤦

cypress/integration/resources.spec.js
```diff
@@ -21,28 +22,28 @@ describe('resources', () => {
...
     cy.location('pathname').should('eq', '/resources/1');
   });
 
-  it('renders the last page of results when paginating to the last page', () => {
+  it('loads the second page of results when using pagination to visit page 2', () => {
+    cy.findByTestId('page 3').click();
+    cy.location('pathname').should('eq', '/resources/3');
+  });
+
+  it('loads the last page of results when paginating to the last page', () => {
```
kylemh, line 46: We can delete this whole test

cypress/integration/resources.spec.js
```diff
@@ -155,49 +141,43 @@ describe('resources', () => {
...
     });
 
-    cy.wait(1000);
-
-    cy.get('[data-test-paid]').each(card => {
-      cy.wrap(card).invoke('attr', 'data-test-paid').should('eq', 'true');
+    cy.findAllByTestId(RESOURCE_CARD).each(card => {
+      cy.wrap(card)
+        .get('[data-test-isPaid]')
+        .invoke('attr', 'data-test-isPaid')
```
kylemh, line 154: What's going on here with `.invoke()`?

I bet we can create a custom Cypress function that let's us check the value of a specific `data-test-*` attribute

https://docs.cypress.io/api/cypress-api/custom-commands.html

`cy.checkCustomDataAttribute('isPaid', false)` would be the usage perhaps?

cypress/integration/resources.spec.js
manthonyg: gone

cypress/integration/resources.spec.js
manthonyg: as far as `invoke()`, from docs:
`cy.get() yields a jQuery object, you can get its attribute by invoking the .attr() method.` 

changed to use different chainers:
`cy.get('[data-test-id="test-example"]')
  .should('have.attr', 'data-test-id', 'test-example')`
 in a Cypress function `checkCustomDataAttribute` as you had suggested

pages/resources/[page].js
```diff
@@ -1,89 +1,297 @@
...
 
 const pageTitle = 'Resources';
 
-function Resources() {
+const ResourceCard = dynamic(() => import('../../components/Cards/ResourceCard/ResourceCard'), {
+  loading: () => <ResourceSkeletonCard numberOfSkeletons={1} />,
+});
+
+Resources.propTypes = {
+  initialValues: shape({
```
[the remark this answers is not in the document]
Comment (manthonyg) [reply pages/resources/[page].js:39]: It works without. I wasn't sure if the type checking mechanism of prop types still benefited us somehow. gone!

cypress/integration/resources.spec.js
manthonyg: changed to `page <page number>` to be a more descriptive testid

pages/resources/[page].js
```diff
@@ -1,89 +1,297 @@
...
+  RESOURCE_RESET_BUTTON,
+} from 'common/constants/testIDs';
 import styles from '../styles/resources.module.css';
 
 const pageTitle = 'Resources';
 
-function Resources() {
+const ResourceCard = dynamic(() => import('../../components/Cards/ResourceCard/ResourceCard'), {
+  loading: () => <ResourceSkeletonCard numberOfSkeletons={1} />,
+});
```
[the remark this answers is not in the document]
Comment (manthonyg) [reply pages/resources/[page].js:36]: I changed it back to the way it was with no dynamic import. I also made it so the skeleton card will flash as long as the inputs are disabled (500ms) because of being submitted- this seems to alleviate the problem aaron noted about not noticing that the resources updated in the UI

manthonyg:
> The last thing we'll wanna add an integration test for is to see that the data is actually changing.
> 
> One way we can do this by collecting some unique meta data for all resources on one page, and then assert that none of that meta data is included in the items on the next page. Prolly the resource name?

After every page load the initial resource names are gathered and made into basically a big paragraph. Then within each of the tests for the different inputs, it compares the current resource names paragraph with the previously gathered ones. I think thatll do

cypress/integration/resources.spec.js
```diff
@@ -3,18 +3,36 @@ import {
...
 
 describe('resources', () => {
-  const CostSelect = 'input#react-select-paid-input';
-  const LanguageSelect = 'input#react-select-languages-input';
-  const CategorySelect = 'input#react-select-category-input';
+  const COST_SELECT = 'input#react-select-paid-input';
+  const LANGUAGES_SELECT = 'input#react-select-languages-input';
+  const CATEGORY_SELECT = 'input#react-select-category-input';
+
+  Cypress.Commands.add('checkCustomDataAttribute', (attribute, value) => {
```
kylemh, line 19: Can you move this to the custom commands file?

The other one looks specific to this file tho, so it's fine here.

cypress/integration/resources.spec.js
```diff
@@ -3,18 +3,36 @@ import {
...
+    cy.get(attribute).should('have.attr', attributeWithoutBrackets, value);
+  });
+
+  Cypress.Commands.add('compareResourceNames', () => {
+    cy.findAllByTestId(RESOURCE_TITLE)
+      .invoke('text')
+      .then($currentResourceNames => {
+        cy.get('@previousResourceNames').should('not.eq', $currentResourceNames);
+      });
+  });
```
kylemh, line 30: So you don't actually need to make this a cypress command. You can just make a function and call it!

cypress/integration/resources.spec.js
manthonyg: Moved it

cypress/integration/resources.spec.js
manthonyg: easy day

PR #1181, opened 2020-07-31
Create column layout for challenge name list

### LINKED ISSUE #1176: OC Challenge Page: Make scrollable list for names added to page
# Feature request
There are currently 50 names displayed on the [OC Challenge page](https://www.operationcode.org/challenge), but it would be beneficial if we had a scrollable list.

## Proposed solution
Make the current list of names on the page scrollable.

Have columns, but they would still need to be scrollable.

justin-reeves:
# Description of changes

Creates a new dynamic column layout for challenger names on the challange page instead of having one continuous list of names. Created to resolve issue 1176
See Slack discussion: https://operation-code.slack.com/archives/C04FASBU9/p1596141056203700
# Issue Resolved

Fixes #1176 

## Screenshots/GIFs

Previous layout had a single column which continued to scroll down page as names are added
image

New layout consists of a dynamic column layout that is based on the number of names desired in each column and flexible to the viewport
(Example Pixel2)
image

(Example LargeScreen)
image

### DIFF
--- a/pages/challenge.js
+++ b/pages/challenge.js
@@ -27,7 +27,7 @@ export const MakeColumnsFromArray = ({ ary }) => {
     content.push(
       <div key={col} className={styles.challengerListColumn}>
         {elements.map(element => (
-          <li key={colPersonStartIndex}>{element}</li>
+          <li key={element}>{element}</li>
         ))}
       </div>,
     );
--- a/pages/challenge.js
+++ b/pages/challenge.js
@@ -12,6 +12,32 @@ const RepoLink = 'https://github.com/OperationCode/front-end/';
 const ChallengeLink = `${RepoLink}blob/main/pages/challenge.js`;
 const CompareLink = `${RepoLink}compare`;
 
+export const MakeColumnsFromArray = ({ ary }) => {
+  const numberNamesPerColumn = 20;
+  const numberCols = Math.floor(ary.length / numberNamesPerColumn) + 1;
+
+  const content = [];
+  let col = 0;
+
+  while (col < numberCols) {
+    const colPersonStartIndex = col * numberNamesPerColumn;
+    const colPersonEndIndex = col * numberNamesPerColumn + numberNamesPerColumn;
+    const elements = challengers.slice(colPersonStartIndex, colPersonEndIndex);
+
+    content.push(
+      <div key={col} className={styles.challengerListColumn}>
+        {elements.map(element => (
+          <li key={colPersonStartIndex}>{element}</li>
+        ))}
+      </div>,
+    );
+
+    col += 1;
+  }
+
+  return content;
+};
+
 function Challenge() {
   return (
     <div className={styles.Challenge}>
@@ -177,10 +203,8 @@ function Challenge() {
             <h6 className={styles.centerText}>
               Here is a list of the people that have completed this before you:
             </h6>
-            <ol>
-              {challengers.map(name => (
-                <li key={name}>{name}</li>
-              ))}
+            <ol className={styles.challengerList}>
+              <Content columns={[<MakeColumnsFromArray ary={challengers} />]} />
             </ol>
           </div>,
         ]}
--- a/pages/challenge.js
+++ b/pages/challenge.js
@@ -4,6 +4,7 @@ import HeroBanner from 'components/HeroBanner/HeroBanner';
 import Content from 'components/Content/Content';
 import OutboundLink from 'components/OutboundLink/OutboundLink';
 import challengers from 'static/operationcode_challenge/names';
+import range from 'lodash/range';
 import styles from './styles/challenge.module.css';
 
 const pageTitle = 'Challenge';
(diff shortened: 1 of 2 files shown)


kylemh:

Built with commit 3a4ef5fdc2ba20f28a38c3c1867814306ed56bca

https://deploy-preview-1181--operation-code-storybook.netlify.app

pages/challenge.js
```diff
@@ -12,6 +12,32 @@ const RepoLink = 'https://github.com/OperationCode/front-end/';
 const ChallengeLink = `${RepoLink}blob/main/pages/challenge.js`;
 const CompareLink = `${RepoLink}compare`;
 
+export const MakeColumnsFromArray = ({ ary }) => {
```
kylemh, line 15: Consider a rename to `NameColumns` because `MakeColumnsFromArray` sounds like a reusable function. 

- It's not reusable (because it has business logic and because it's impure - this sounds bad/negative, but there's no such connotation - it just is what it is).

- While all JSX elements are functions, we typically don't name them like functions because they're supposed to _look_ like structured content. For example, nobody write a CreateButton component (`<CreateButton>`)... well, I guess they could, but only if it's a button that creates something when clicked! I hope you get my point.

pages/challenge.js
```diff
@@ -12,6 +12,32 @@ const RepoLink = 'https://github.com/OperationCode/front-end/';
...
+  const numberNamesPerColumn = 20;
+  const numberCols = Math.floor(ary.length / numberNamesPerColumn) + 1;
```
kylemh, line 17: Since you're using `challengers` (a variable defined outside the scope of this method), you've made an impure function (this sounds bad/negative, but there's no such connotation - it just is what it is). 

Since you've already made an impure, you can ditch the argument entirely, and just use `challengers.length`.

pages/challenge.js
```diff
@@ -12,6 +12,32 @@ const RepoLink = 'https://github.com/OperationCode/front-end/';
...
+        {elements.map(element => (
+          <li key={element}>{element}</li>
+        ))}
+      </div>,
+    );
+
+    col += 1;
+  }
+
+  return content;
```
kylemh, line 38: You don't need to rewrite this. Mainly, I just enjoy showing people how to immutably construct mungings (no using `let` / `[].push()`). I would be happy if you took some of my variable renaming ideas though!

```jsx
  import range from 'lodash/range';
  
  //
  
  const numberOfNamesPerColumn = 20;
  const numberOfColumns = Math.floor(names.length / numberOfNamesPerColumn) + 1;
  const columns = range(1, numberOfColumns);

  const content = columns.map((columnNumber, index) => {
    const startIndex = index * numberOfNamesPerColumn;
    const endIndex = columnNumber * numberOfNamesPerColumn;
    const namesInColumn = challengers.slice(startIndex, endIndex);
    
    return (
      <ol key={columnNumber} className={styles.challengerListColumn}>
        {namesInColumn.map((name) => (
          <li key={name}>{name}</li>
        ))}
      </ol>,
    );
  });

  return content;
```

pages/challenge.js
```diff
@@ -12,6 +12,32 @@ const RepoLink = 'https://github.com/OperationCode/front-end/';
...
+  const content = [];
+  let col = 0;
+
+  while (col < numberCols) {
+    const colPersonStartIndex = col * numberNamesPerColumn;
+    const colPersonEndIndex = col * numberNamesPerColumn + numberNamesPerColumn;
+    const elements = challengers.slice(colPersonStartIndex, colPersonEndIndex);
+
+    content.push(
+      <div key={col} className={styles.challengerListColumn}>
```
kylemh, line 28: this should be `ol`. `li` can only ever be children of `ol` elements and `ol` element should only ever have `li` elements as immediate children.

See here for more: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol

pages/challenge.js
```diff
@@ -177,10 +203,8 @@ function Challenge() {
             <h6 className={styles.centerText}>
               Here is a list of the people that have completed this before you:
             </h6>
-            <ol>
-              {challengers.map(name => (
-                <li key={name}>{name}</li>
-              ))}
+            <ol className={styles.challengerList}>
```
kylemh, line 206: You can delete this and the CSS class entirely. I don't think there will be style issues.

justin-reeves:
Unintentionally closed PR. Reopening!

pages/styles/challenge.module.css
```diff
@@ -22,3 +22,15 @@
...
+
+.Challenge .challengerListColumn {
+  padding: 0;
+  align-self: flex-start;
+  flex-basis: 225px;
+  list-style-position: inside;
+}
+
+.Challenge .challengerListContainer div {
+  width: 99vw;
```
kylemh, line 34: Hmmm this wont work cuz - as you can see - on mobile views there's a horizontal scrollbar. What you'll actually wann do is pass a className to `Content` which does `width: 100%` and then apply that same property to the empty div on line 199 + `.challengerListContainer`

kylemh: Screen Shot 2020-07-31 at 1 57 59 PM
like so

justin-reeves: As we discussed in Slack, the Content component className is not exposed so for now the fix will be to just lower width to 85vw. In the future we can expose the className for the Content component in order to change width to 100% as mentioned.

PR #1194, opened 2020-08-11
Unit tests for user registration page and validation changes

justin-reeves:
# Description of changes

This PR introduces a number of changes related to user registration validation as well as accompanying cypress unit tests.
Some of the major changes include:

1. Unit tests for each field of the join page (Email, password, names, zipcode).
2. Unit tests for the isMinPasswordStrength() utility function.
3. Alterations to the passwordStrengthRegex to include a minimum character count (8).
4. Overhaul of the mockPassword() function to utilize more faker functionality as well as allow for more flexible use in the future.

See commit messages for more detailed additions/fixes

# Issue Resolved

Fixes #


kylemh:

Built with commit e40e7100e2b106fb0e7d54677540c7589ded12aa

https://deploy-preview-1194--operation-code-storybook.netlify.app

common/constants/validations.js
```diff
@@ -1,2 +1,2 @@
 export const minimumPasswordLength = 8;
-export const passwordStrengthRegex = /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d).+/;
+export const passwordStrengthRegex = /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d).{8,}/;
```
kylemh, line 2: this isn't necessary, because we test length via the variable above (connected to `Yup.min()`).

common/utils/__tests__/validator-utils.test.js
```diff
@@ -1,63 +1,56 @@
...
-      expect(isValidZipcode('testString')).toStrictEqual(true);
-    });
-
-    it('should return false when passed an empty string', () => {
-      expect(isValidZipcode('')).toStrictEqual(false);
-    });
-
-    it('should return false when passed undefined', () => {
-      expect(isValidZipcode()).toStrictEqual(false);
-    });
```
kylemh, line 15: why remove these?

cypress/integration/join.spec.js
```diff
@@ -1,7 +1,7 @@
 import { validationErrorMessages } from '../../common/constants/messages';
```
kylemh, line 1: I know you asked about this, and I feel like I've misled you. For this file, while we do want to ensure that the errors render, I also still think there's value in ensuring that a user doesn't navigate past `/join` when using an incorrect password. The routing aspect is what makes it a valuable integration test.

test-utils/mockGenerators/mockPassword.js
```diff
@@ -1,15 +1,48 @@
...
+ * @param {boolean} [hasOneLowercaseChar=true] Password should contain a lowercase char
+ * @param {boolean} [hasOneUppercaseChar=true] Password should contain an uppercase char
+ * @param {boolean} [hasOneNumber=true] Password should contain a number
  * @returns {string}
  */
-export default function mockPassword(isValid = true) {
-  return isValid
-    ? `${faker.internet.password(minimumPasswordLength)}!1Aa${faker.internet.password(1)}`
-    : `${faker.internet.password(minimumPasswordLength - 2)}`;
+export default function mockPassword({
```
kylemh, line 18: nice!

common/utils/validator-utils.js
```diff
@@ -1,24 +1,11 @@
 import { passwordStrengthRegex } from '../constants/validations';
 
-/**
- * Take a string and determine if it has non-whitespace characters in it
- *
- * This may seem counter-intuitive, but it's difficult to get a zipcode regex correctly.
- * See https://stackoverflow.com/questions/578406/what-is-the-ultimate-postal-code-and-zip-regex
- * for more info
- *
```
kylemh, line 9: I would leave `isValidZipcode` alone if for no other reason than because of these comments. If it's removed, I'm sure we'll have people popping up asking about our lack of validation. While this function doesn't do much (it simply prevents whitespace submissions), it exists and says to most: "we have some sort of validation on zipcodes" and it explains why it's weak for anybody that dives deeper.

common/utils/__tests__/validator-utils.test.js
justin-reeves: `isValidZipCode()` only checks to ensure that the user doesn't enter an empty string or string of just spaces. I added the `trim()` to the Yup call on zipcode to check this just as we did with `firstName` and `lastName`. There is no need for `isValidZipCode()` and therefore no need for the unit tests.

cypress/integration/join.spec.js
justin-reeves: I see. So we want to ensure that the errors render but also that the front-end doesn't actually allow a submission regardless of errors?

common/utils/validator-utils.js
justin-reeves: The validation can be accomplished via `Yup.string().trim()`. I can add it back though if you want!

common/constants/validations.js
justin-reeves: I had taken the Yup.min() check off as it could all be taken care of via the regex. However, I can put it back on!

common/utils/validator-utils.js
justin-reeves: The validation can be accomplished via Yup.string().trim(). I can add it back though if you want!

kylemh: I guess that's fair then, please move the comments over too

common/constants/validations.js
kylemh: Yes please! I'd prefer it the other way around

common/utils/__tests__/validator-utils.test.js
```diff
@@ -1,63 +1,56 @@
...
-    it('should work with a lowercase letter, an uppercase letter, and a number', () => {
-      const password = 'aA1';
-      expect(isMinPasswordStrength(password)).toStrictEqual(true);
-    });
+    it(
+      'should work with 8 characters, ' +
+        'a lowercase letter, ' +
+        'an uppercase letter, ' +
+        'and a number',
+      () => {
```
kylemh, line 11: great use of a test table here

components/RegistrationForm/__tests__/RegistrationForm.test.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 import { fireEvent, render, waitFor } from '@testing-library/react';
-import { networkErrorMessages, validationErrorMessages } from 'common/constants/messages';
+import { networkErrorMessages } from 'common/constants/messages';
 import createSnapshotTest from 'test-utils/createSnapshotTest';
 import mockUser from 'test-utils/mockGenerators/mockUser';
-import mockPassword from 'test-utils/mockGenerators/mockPassword';
+/* import mockPassword from 'test-utils/mockGenerators/mockPassword'; */
```
kylemh, line 6: commented out?

cypress/integration/join.spec.js
```diff
@@ -110,134 +165,403 @@ describe('join', () => {
...
+    cy.get(inputFields.zipcode).type(validUser.zipcode);
+    cy.get('button[type="submit"]').click();
+
+    cy.url().should('contain', '/join');
+
+    cy.get('div[role="alert"]')
+      .should('have.length', 1)
+      .should('contain', validationErrorMessages.required);
+
+    cy.getCookies().should('have.length', 0);
```
kylemh, line 469: It looks like you re-use this block of logic many times.

Maybe you could make a function like:

```js
const assertFailedLogin = ({ numberOfErrors }) => {
  cy.url().should('contain', '/join');
  cy.getCookies().should('have.length', 0);

  cy.get('div[role="alert"]')
    .should('have.length', numberOfValidationErrors)
    .should('contain', validationErrorMessages.required);
}

// ...

assertFailedLogin({ numberOfErrors: 1 });
```

kylemh, line 469: It looks like you re-use this block of logic many times.

Maybe you could make a function like:

```js
const assertFailedLogin = ({ numberOfErrors }) => {
  cy.get('button[type="submit"]').click();

  cy.url().should('contain', '/join');
  cy.getCookies().should('have.length', 0);

  cy.get('div[role="alert"]')
    .should('have.length', numberOfValidationErrors)
    .should('contain', validationErrorMessages.required);
};

// ...

assertFailedLogin({ numberOfErrors: 1 });
```

kylemh: You could also group all of the failed submission test cases into one `describe` block and put the above into an `afterEach()`

components/RegistrationForm/__tests__/RegistrationForm.test.js
justin-reeves: Was making sure I didn't need it while re-writing `mockPassword()`. Will remove it now.

cypress/integration/join.spec.js
justin-reeves: Sounds like a good idea! I'll work that into the solution

common/constants/validations.js
```diff
@@ -1,2 +1,2 @@
 export const minimumPasswordLength = 8;
-export const passwordStrengthRegex = /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d).{8,}/;
+export const passwordStrengthRegex = /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d).+/;
```
kylemh: can you rename this one to `requiredCharactersRegex` per our chat?

justin-reeves: Yup sorry, thought we only meant the function but that makes sense too!

PR #1217, opened 2020-09-15
Removed stored user PII in cookies

### LINKED ISSUE #1146: PII is stored in browser cookies
# Bug Report
Once logged in, a user's first name, last name, and zip code are being tracked in the browser cookie storage.

Instead, wherever we need this info we should make an API request using the `getUserPromise` method. We only need to store the token in the cookie (for now - this, too, is changed eventually).

I logged in, went to inspect element > storage > cookies for current web session.

ajpgross:
# Description of changes
1. Removed set and remove of the user info cookies in cookie-utils. JWT is still set/removed in util functions.
2. Cleaned up login functions by removing references to user object.
3. Updated e2e tests for removal of PII from cookies; included a check of the decoded JWT cookie instead of the old user info cookies.

# Issue Resolved
Fixes #1146

## Screenshots/GIFs
Before:
image
After:
image

### DIFF
--- a/common/utils/auth-utils.js
+++ b/common/utils/auth-utils.js
@@ -2,8 +2,8 @@ import Router from 'next/router';
 import nextCookie from 'next-cookies';
 import { setAuthCookies, removeAuthCookies, hasValidAuthToken } from './cookie-utils';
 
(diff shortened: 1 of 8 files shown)


kylemh (approved):
per-fect

PR #1218, opened 2020-09-19
Upgrade Storybook to v6

justin-reeves:
# Description of changes

Storybook migration from version 5 to version 6 as well as Story upgrade to CSF.

There are issues with the testing however. This can be resolved in a subsequent PR.
Additional comments will be added to story args for better documentation as well.

# Issue Resolved

Fixes #NA


kylemh:
We've removed Storyshots which will lower coverage, but we'll soon be adding Chromatic which should theoretically revive the coverage lost in this PR.

PR #1262, opened 2020-10-02
chore: update formik to v2

### LINKED ISSUE #1236: Upgrade Formik to v2
# Feature request
We're using `formik@1^`. I wanna use `formik@2^`.

## Proposed solution
- Upgrade `formik` to the latest version 
- Convert implementations of formik to leverage `useFormik` hook instead of `connect()` or `<Formik>` render props.

Tomekmularczyk:
# Description of changes
Updates Formik dep to v2 following this guide https://formik.org/docs/migrating-v2

# Issue Resolved

Fixes #1236


pages/resources/[page].js
```diff
@@ -260,6 +257,7 @@ function Resources() {
                     </Button>
 
                     <Button
+                      onClick={() => handleReset(resetForm)}
```
Comment (Tomekmularczyk) [inline pages/resources/[page].js:260]: I know this one is not ideal, but `onReset` for some reason was fired every time a component would re-render, not only when pressing the reset button itself...

Tomekmularczyk:
@kylemh could you help me with this one? there is one cypress test failing. It seems to me that it's rather some data not returned rather than UI problem 🤔

pages/resources/[page].js
Comment (kylemh) [reply pages/resources/[page].js:260]: One thing I bet we could do is instead leverage the `onReset` prop on the Form component. Note how we dont have an `onClick` on the submit button?

Comment (Tomekmularczyk) [reply pages/resources/[page].js:260]: yes, this is how it used to be bofore - "onReset prop on the Form component" but after updating Formik to v2, onReset is being called not only when you click the "Reset". For example, it's also triggered when you hit "Search" making the whole form reset...
Maybe I need to investigate it once more...

Comment (kylemh) [reply pages/resources/[page].js:260]: Hmmm I'd definitely like a better understanding, and would appreciate any investigation.

Tomekmularczyk:
@kylemh one question though, do you know whether e2e tests failing is a common thing, or did I definitely broke something?

pages/resources/[page].js
```diff
@@ -194,11 +194,8 @@ function Resources() {
                 handleSubmit(values, actions);
                 actions.setSubmitting(true);
               }}
-              onReset={(values, actions) => {
-                handleReset(values, actions);
-              }}
             >
-              {({ isSubmitting }) => (
+              {({ isSubmitting, resetForm }) => (
                 <Form role="search">
```
Comment (kylemh) [inline pages/resources/[page].js:199]:
```suggestion
                <Form role="search" onReset={resetForm}>
```

to be clear, you tried this and it didnt work?

kylemh:
You didn't break anything. I was testing it out and it appears we have an issue with rate limiting. I've conveyed the issue to the engineer in charge of the resources microservice.

I'd like to wait until that issue is resolved before merging this (so that we can ensure tests pass).

pages/resources/[page].js
Comment (Tomekmularczyk) [reply pages/resources/[page].js:199]: yes, it led to Maximum callstack exceeded error. 
* I would click "Search" - component would re-render, leading to onReset being triggered (for no reason?) and onReset would reset the form, which would lead some re-render and so on...
* the same when you clicked "Reset" - it would lead to endless loop with `onReset` being called on and on

Comment (Tomekmularczyk) [reply pages/resources/[page].js:199]: I one thought I have is that when you do `formikBag.resetForm()` the side effect would `onReset` being triggered, which I think was not the initial intention but could be a side-effect that's causing this

Comment (Tomekmularczyk) [reply pages/resources/[page].js:199]: @kylemh I guess I'm right here, take a look at the implementation of [`resetForm`](https://github.com/formium/formik/blob/master/packages/formik/src/Formik.tsx#L378) there is a condition `if(props.onReset)` inside which calls `onReset` callback.
So if we have a `onReset` calling `resetForm` it will lead to endless loop.

Comment (kylemh) [reply pages/resources/[page].js:199]: AH! https://github.com/Tomekmularczyk/front-end/blob/chore/update-formik/components/Form/Form.js

Looks like we may need to edit some stuff here then

Comment (Tomekmularczyk) [reply pages/resources/[page].js:199]: I think it's fine, the props are correct and `onReset` callback was only used in this file alone

Comment (kylemh) [reply pages/resources/[page].js:199]: I think it's because it's the only form with a reset button.

Comment (kylemh) [reply pages/resources/[page].js:199]: What if we have _no_ callback for reset handler (including the button) and just let it by `type="reset"`.

Comment (Tomekmularczyk) [reply pages/resources/[page].js:199]: Just checked, nothing will happen. If we want to keep `onReset` we need to find a way not to call `resetForm` in it

Comment (kylemh) [reply pages/resources/[page].js:199]: In the original issue, I think I referenced getting rid of <Formik> usages, didn't I? I feel like if we're leveraging the `useFormik` hook we may entirely ditch the `Form` component, no?

Comment (Tomekmularczyk) [reply pages/resources/[page].js:199]: true :D just noticed that, though I think updating the dependency and refactoring all forms could be separate issues 😅  seems like a lot of work

Comment (Tomekmularczyk) [reply pages/resources/[page].js:199]: What is more `onReset` is being called even when the form is reinitialized so

* select something in the form 
* press submit
* URL is updated
* component re-renders and initialValues change
* Formik reinitializes because of new initialValues and `enableReinitialize` prop
* onReset is being called because of that reinitialization
* `router.push(pathname, '/resources/1', { shallow: true });` is called and URL resets
* because URL is different again, initialValues change again and form is reinitialized to initial - empty - state

Seems like `onReset` has a very different purpose now than it had in v1. IMO, I would remove it.

@kylemh let me know if there's anything I could/should do with this because I don't think there's much more I can do :)

Comment (kylemh) [reply pages/resources/[page].js:199]: it sounds like future PRs to refactor the forms would be best

Comment (kylemh) [reply pages/resources/[page].js:199]: once the resources API is back up and running I'll merge this

if you wanna do any of what we discussed above, feel free to branch off of this

Comment (Tomekmularczyk) [reply pages/resources/[page].js:199]: thank you a lot :) for now, I think I made my share of work in this repo, I will leave refactoring to hooks for other Hacktoberfest participants :)

kylemh:

PR #1263, opened 2020-10-03
feat(#1141): create <SurveyForm /> and create public route /survey

### LINKED ISSUE #1141: Duplicate current registration form and move it to a new, public route
# Feature request
We want to create a user survey that looks a lot like registration in the information it gathers. It will eventually be accessed via a slash command on Slack. 

## Proposed solution
- Duplicate current registration form 
- move it to a new, public route called “survey”

## Additional context
Related to operationcode-pybot issue #184 and backend issue #347

hackily:
# Description of changes
Creates a new `<SurveyForm />` component and a new page `/survey` that consumes the component. Both password fields and the email confirmation field have been removed - presumably, those won't be necessary in a survey.

There is no backend api for accepting survey form data so I left one comment in the `handleSubmit()` and a couple comments in the test file, where the test should be changed in the future to assert that the api call is being made.

# Issue Resolved

Fixes #1141

## Screenshots/GIFs

image

### DIFF
--- a/components/SurveyForm/SurveyForm.js
+++ b/components/SurveyForm/SurveyForm.js
@@ -1,0 +1,145 @@
+import React, { useState } from 'react';
+import { func, number, oneOfType, shape, string } from 'prop-types';
+import { Formik, Field } from 'formik';
+import * as Yup from 'yup';
+import { getServerErrorMessage } from 'common/utils/api-utils';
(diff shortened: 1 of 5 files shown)


kylemh:
@hackily - really sorry that I had you work on something were not acting on.

I've closed the related issue and you can still consider this PR "merged", despite the fact that it isn't.

PR #1265, opened 2020-10-05
Fix - FAQ page volunteer section accordion components aren't full width

michaeljoy255:
# Description of changes
Added 100% width to the columnsContainer CSS class so it uses the entire width available within it's parent's element. The fix did not appear to impact any of the other uses of the columnsContainer throughout the website.

# Issue Resolved
No issue #, just an unsolicited PR :)

The columnsContainer CSS class may not work as intended if there are not enough elements within it to expand it to it's parent element's width.

## Screenshots
Before change:

before

After change:

after


kylemh (changes requested):
@michaeljoy255 this would affect every `Content` component usage on every page, not just FAQ.

I like the problem you were trying to solve, but I think you'd need to change code in `pages/faq.js` and `pages/styles/faq/module.css` instead.

michaeljoy255:
FAQ uses the Content component and Accordion component. The Content component internally uses the columnsContainer class I want to adjust, but on the FAQ page only. I'm unsure how to achieve this since it seems like React does some scoping with the CSS modules that I'm not quite understanding (I come from a Vue background). Not sure how I can use the style.prop syntax to apply a class in the FAQ page that will effect columnsContainer if I can't apply the class directly to the element.

kylemh:
You're totally correct in that this page uses those components; however, MANY pages use this component - it's meant to be reusable. So a change here, affects all of those usages too. For example, [on your preview deployment's code schools page](https://operation-code-git-ae1bd3f32eae3acc0ba1fce3c6f874326b67485c.operation-code.vercel.app/code_schools), notice how the introductory text is all in one column when it used to be spread across 3 columns?

The CSS scoping has nothing to do with React and all about css modules (which the Vue CLI actually scaffolds and uses under the hood as well) and how Next.js integrates it with the folder structure and file naming (a style file only uses CSS Modules if suffixed with a `${fileName}.module.(css|sass|scss|less)`).

Info:
https://github.com/css-modules/css-modules
https://nextjs.org/docs/basic-features/built-in-css-support

Unfortunately, this fix may be bigger than I initially thought! It seems that the `Accordion` component doesn't have a great story with the `Content` component. The only other place Accordion is used is in `pages/resources/[page].js`, and while it's _technically_ a child of a `Content` component, there's custom styling applied to a parent which hard-codes max-width as `70vw` meaning the width of the Accordion goes from there.

I think the quickest fix would be to also hard-code widths and use media queries to adjust as needed. The best fix would be to give the `Content` component a prop that can be used in conjunction with [classnames](https://github.com/JedWatson/classnames) to control/alter the styling of the div that has the `.columnsContainer` class, like so:

```
<div className={classNames(styles.columnsContainer, columnContainerClassName)}>
```

Once that's done, we can just pass custom stylings into our `<Content>` usages on `pages/faq.js`.

kylemh (approved):

kylemh:

Issue #1142, opened 2020-06-27
Create new form on front-end to trigger user connection process 


ksmacleod99:
# Feature request
Create a new form to trigger user connection process between Slack and OC profiles

## Proposed solution
Create a new form to trigger user connection process between Slack and OC profiles

## Additional context
Related to backend issue #346

ksmacleod99:
@kylemh in your mind, will this need a design?

kylemh:
No. The form should have 4 inputs in two groupings.

Your OperationCode.org email and your Slack email

Beissner:
What's the status on this issue? Are we waiting for backend issue #346 to be completed first?

kylemh:
Closing because we won't be doing the OAuth work.

Issue #1124, opened 2020-05-24
Redesign Join page


mikerod-sd:
# Feature request

Add wording on https://operationcode.org/join that explains why we collect demographic information

Also add wording on next page after submit (or whatever page is returned after final submit) that outlines how to join slack (i.e. "Thanks for submitting your information, please be on the lookout for an email which will have a link to join our slack")

## Proposed solution

See above.

## Potential alternative solutions

N/A

## Additional context

Per discussion in the mod channel on slack.

Too-Far:
I can tackle this if you guys want! I should be able to do it Tuesday night if not sooner.

datphotogurl:
Adding continued functionality requests to the https://operationcode.org/join page:

1. Add checkbox and new window expansion for "I have read the Operation Code's _**Code of Conduct**_ and agree to abide by it during my time as a member."
2. Add checkbox and new window expansion of rules "I agree to adhere to _**Slack's Community Guidelines**_"

Screenshot of La Techqueria's Join page is attached for an example.
<img width="670" alt="Screen Shot 2020-07-30 at 3 40 03 PM" src="https://user-images.githubusercontent.com/51416206/88967000-31690d00-d27b-11ea-864b-4ab6805a3822.png">

kylemh:
No back-end needed. We'll just do front-end validation regarding these checks since they're not legal in nature.

kylemh:
@datphotogurl can we get links to the two policies? I'd consider this ready to be worked on if we have those links (assuming we'll copy... their copy)

datphotogurl:
> @datphotogurl can we get links to the two policies? I'd consider this ready to be worked on if we have those links (assuming we'll copy... their copy)

Hi @kylemh, I was walking through this issue with Lianis as an exercise, but yes, I can add the two sets of content. Currently, the Code of Conduct is being updated by the Board and Mods. 

The link for the Code of Conduct: https://github.com/OperationCode/operationcode_docs/blob/master/community/code_of_conduct.md 

We do not currently have a repo for a Slack Community Guideline but I can write up that content and provide that to you to either embed in Github or the website. Which do you suggest?

kylemh:
I would just make a new markdown file in that same repo!

kylemh:
@datphotogurl can we get agreed upon content for this?

datphotogurl:
Yes, I need to work on it on the weekend as my weekday schedule is ramped
up now.

*Cynthia S. Kao,*
[...]

PR #1268, opened 2020-10-05
Issue #1142

ksmacleod99:
# Description of changes
**Need help first** Will satisfy #1124 

# Issue Resolved
Added checkbox, labels, and appropriate links.

Need assistance with: 
- Validation
- tests
- ideas for styling input labels to match checkbox labels or vice-versa

## Screenshots/GIFs
Screen Shot 2020-10-05 at 7 16 09 PM


kylemh:
DMed ya ❤️

pages/profile/index.js
```diff
@@ -8,7 +8,7 @@ import withAuthSync from 'decorators/withAuthSync/withAuthSync';
 import { getUserPromise } from 'common/constants/api';
 import { capitalize, startCase, toLower } from 'lodash';
 import { format } from 'date-fns';
-import styles from 'styles/profile.module.css';
+import styles from '../styles/profile.module.css';
```
kylemh, line 11: 👀

ksmacleod99:
Getting this error when pushing. I'm sure one of the CI tests will give more details. 
I will need some help putting the tests together. 

Screen Shot 2020-10-13 at 10 20 27 AM

components/Form/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,80 @@
...
+            type="checkbox"
+            value={value || ''}
+          />
+        </div>
+
+        <div className={styles.floatRight}>
+          <Label for={name} isHidden={isLabelHidden} className={styles.labelAfterCheckbox}>
+            {label}
+          </Label>
+        </div>
```
kylemh, line 64: We should have the label component be listed before the input component for accessibility purposes, but then we can group the input and label into one `div` element that uses `display: flex;` and `flex-direction: row-reverse;` to make them _appear_ to render in the "correct" order (checkbox first, then label).

components/Form/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,80 @@
...
+
+        <ErrorMessage name={name}>
+          {message => {
+            return hasErrors ? (
+              <Alert className={styles.errorMessage} data-testid={CHECKBOX_ERROR} type="error">
+                {message}
+              </Alert>
+            ) : null;
+          }}
+        </ErrorMessage>
```
kylemh, line 74: Could we adjust the styling of the error message (or the container that holds it - whatever you need) so that it renders below each checkbox? It looks a little off rendering to the side.

components/Form/Checkbox/__stories__/Checkbox.stories.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
kylemh, line 1:
```jsx
import React from 'react';

import Checkbox from '../Checkbox';

export default {
  component: Checkbox,
  title: 'Form/Checkbox',
};

const Template = arguments_ => (
  <>
    <span>
      NOTE: This component&apos;s story has no context outside of Formik and will not function
      properly
    </span>
    <Checkbox {...arguments_} />
  </>
);

const fieldName = 'someCheckbox';

export const Default = Template.bind({});

/** Default Checkbox supplied with only required args */
Default.args = {
  field: {
    name: fieldName,
  },
  form: {
    touched: { [fieldName]: false },
    errors: { [fieldName]: '' },
  },
  label: 'Checkbox label',
};

```

components/Form/Checkbox/__tests__/Checkbox.test.js
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
```
kylemh, line 1: You can delete this file. We'll get plenty of coverage from Cypress + the RegistrationForm jest test.

components/Form/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,80 @@
...
+  field: shape({
+    name: string.isRequired,
+  }).isRequired,
+  form: shape({
+    touched: object.isRequired,
+    errors: objectOf(string),
+  }).isRequired,
+  isLabelHidden: bool,
+  id: oneOfType([string, number]),
+  label: string.isRequired,
```
kylemh, line 20:
```suggestion
  label: oneOfType([node, string]).isRequired,
```

components/RegistrationForm/RegistrationForm.js
```diff
@@ -175,6 +182,20 @@ function RegistrationForm({ initialValues, onSuccess }) {
...
+            <Field
+              name="codeOfConduct"
+              label="I have read the Code of Conduct"
+              component={Checkbox}
+              disabled={isSubmitting}
+            />
+
+            <Field
+              name="communityGuidelines"
+              label="I have read the Community Guideliens"
```
kylemh, line 195: Assuming you allowed for node type via prop-types...

```suggestion
              label={
                <>
                  I have read the{' '}
                  <OutboundLink
                    hasIcon={false}
                    href="some-url.com"
                    analyticsEventLabel="Looked at Code of Conduct"
                  >
                    Code of Conduct
                  </OutboundLink>
                </>
              }
```

pages/profile/index.js
kylemh:
```suggestion
import styles from 'styles/profile.module.css';
```

kylemh (commented):
Really good work here! Just a few tid-bits.

In the RegistrationForm unit test, you'll want to adjust line 19 to:
`const user = { ...mockUser(), codeOfConduct: true, communityGuidelines: true };`

PR #1275, opened 2020-10-06
added google's Lighthouse CI check

### LINKED ISSUE #1230: Use Google's Lighthouse via a CI check
# Feature request
Google's [Lighthouse](https://developers.google.com/web/tools/lighthouse) is a tool used by developers to get a simple measurement on a page's performance, accessibility score, and SEO rating. We should use it via some CI step (CircleCI, GitHub Actions, something else).

## Proposed solution
Leverage Lighthouse in this repository's continuous integration against the landing page in a way that lets me confirm that we haven't strayed to far from the scores that https://www.operationcode.org/ is receiving.

## Additional context
A quick search yields me many potential tools:
- [Lighthouse CI Action](https://github.com/marketplace/actions/lighthouse-ci-action)
- [Lighthouse Check](https://github.com/marketplace/actions/lighthouse-check)
- [Web Performance Audits With Ligthhouse](https://github.com/marketplace/actions/web-performance-audits-with-lighthouse)
- [Orbs](https://circleci.com/developer/orbs?query=lighthouse)

I _think_ I'd prefer GitHub action so that we're parallelizing CI tasks; however, I'm open to using Orbs or even some other solution. I'll leave the evaluation of the best tool up to you.

BhuwanChandra:
# Description of changes

added lighthouse ci check to get a measurement on the performance of website.
# Issue Resolved

Fixes #1230


.github/workflows/lighthouse.yml
```diff
@@ -0,0 +1,52 @@
+name: lighthouse
+on:
+  push:
+    branches-ignore:
+      - master
```
kylemh, line 5: do we want to ignore master?

How does it know what to compare against

BhuwanChandra: I thought we have to add this for other branches so that whenever we merge a branch into master it should be at recommended performance scores. Should I need to change it to all branches?

.github/workflows/lighthouse.yml
```diff
@@ -0,0 +1,52 @@
...
+            ${{ runner.os }}-yarn-
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile --non-interactive
+
+      - name: Audit URLs using Lighthouse
+        uses: treosh/lighthouse-ci-action@v3
+        with:
+          urls: |
+            https://operationcode.org/
```
BhuwanChandra, line 45: It will compare against the urls we have specified here in urls field.

kylemh: ah okay - willing to try it! Any reason there's no feedback in _this_ PR from Lighthouse?

BhuwanChandra: Its because I have added the lighthouse check on push not on pull requests. I think, I also need to add pull requests into it.

BhuwanChandra: I have made some changes, now it's showing the lighthouse check as well.

kylemh:
I see

>   No GitHub token set, skipping GitHub status check.

in the workflow spot.

Screen Shot 2020-10-08 at 9 20 13 AM

Do we need to do something about that?

Additionally, I'm still not sure what to expect from this check. What would make it fail?

BhuwanChandra:
> A status check, if configured, is a message that appears on every PR and typically includes information such as the results of a test or the success of a build.

You need to set lighthouse github app in this repository to get the status checks and performance scores on every pull requests.
Capture
something like this.

to set github status check:
1. go to https://github.com/apps/lighthouse-ci
2. add lighthouse CI to this repository
3. add LHCI_GITHUB_APP_TOKEN to the secrets of the repository.

this is totally optional, otherwise we need to go to the details each time to check the performance scores.

basically in these checks, we can run our lighthouse checks against some predefined performance scores (we need add them into our `lighthouserc.js` file) and if the current scores are below those predefined scores then this check will get failed. 

currently I haven't defined any scores yet, If you want me add the scores let me know I will add some performance scores.

and also let me know if you have enabled the github status check then I need to make some changes in `lighthouse.yaml` file in order to make it work.

kylemh:
Awesome. Added the token: `LHCI_GITHUB_APP_TOKEN` to GitHub. Updated your branch! Let's see if it works magically.

BhuwanChandra:
I have made the required changes but still it is showing `No GitHub token set, skipping GitHub status check` this, can you check if  `LHCI_GITHUB_APP_TOKEN`  is in the secrets of the repository or not?
I have added some minimum scores for each checks and now the checks will get failed if the performance scores are below those scores.

kylemh:
@BhuwanChandra I think it was an annoying typo issue. The actual GitHub app refers to the would-be token as: `LHCI_GITHUB_APP_TOKEN`, but this workflow expects `LHCI_GITHUB_TOKEN` 🙃  https://github.com/treosh/lighthouse-ci-action/issues/60#issuecomment-684040439

kylemh:
Hmmmm, not sure what else could be going on.

I have the token added via secrets under both monikers - just in case:

Screen Shot 2020-10-14 at 5 44 06 PM

I've had them in since you began working on this - ignore the time stamp - I just moved them to the org level instead of just this repository.

kylemh:
Do you want to try this one out https://github.com/OskarAhl/Lighthouse-github-action-comment/blob/main/.github/workflows/lighthouse-on-vercel-preview-url.yml

BhuwanChandra:
Ok, I will try this one!

Issue #1277, opened 2020-10-06
Footer links dont open when clicked on iOS mobile


kylemh:
# Bug Report
## What is the current behavior?
Clicking any footer link does nothing.

## What is the expected behavior?
Clicking any footer link should navigate and scroll to the top of the navigated page.

## What steps did you take to get this behavior?
Used my iPhone and cried at the realization.

### Operating System
iOS 14

### Browser
Chrome iOS (uses the same rendering engine as Safari, so either should be the same)

rzhekova:
I can give this one a try if that's ok

kylemh:
Go for it

rzhekova:
Hey @kylemh I think I've found the culprit, I'll send a PR over to you in a bit

kylemh:
Closed in https://github.com/OperationCode/front-end/pull/1280

PR #1280, opened 2020-10-07
PR for "Footer links dont open when clicked on iOS mobile #1277"

rzhekova:
This PR resolves issue #1277.

Hello, it seemed that the issue was related to Nextjs and it was happening on desktop Safari too. As I was investigating the cause, I noticed that this `SecurityError: Attempt to use history.replaceState() more than 100 times per 30 seconds` error was happening as I scrolled the page, following that I'd not be able to use some of the footer links and this other error `Unhandled Promise Rejection: SecurityError: Attempt to use history.pushState() more than 100 times per 30 seconds` would start firing on click.

I was able to fix the problem by setting `scrollRestoration` to `false` (I can also get rid of it altogether if you want). Let me know what you think!

image

image

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -13,7 +13,7 @@ const nextConfig = withBundleAnalyzer({
 
   experimental: {
     productionBrowserSourceMaps: true,
-    scrollRestoration: true,
+    scrollRestoration: false,
   },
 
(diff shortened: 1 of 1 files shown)


kylemh (approved):
Amazing investigative work ❤️  - I committed a comment, and we're good to go!

PR #1283, opened 2020-10-09
enable upvote/downvote feature on resources

chynh:
# Description of changes

Enable Upvote/Downvote feature on a resource

# Issue Resolved

Fixes #NA

## Screenshots/GIFs

upvote:downvote

## Note
There's an auth issue on the resources API side that needs to be resolved first. I am seeing that `authorization` header is being set properly when making a request but was getting a `401` response. I was only able to verify that this is working by hardcoding in `x-apikey` header.

Screen Shot 2020-10-08 at 11 02 44 PM
Screen Shot 2020-10-08 at 11 02 58 PM


common/constants/api.js
```diff
@@ -52,3 +52,7 @@ export const updateUser = userInfo => {
     ...formatUserData(userInfo),
   }).then(({ data }) => data);
 };
+
+/* PUT REQUESTS */
+export const updateResourceVoteCount = ({ id, voteDirection }) =>
+  put(`api/v1/resources/${id}/${voteDirection}`);
```
kylemh, line 58: Interesting that this isn't simply a POST. Any explanation?

components/Cards/ResourceCard/ResourceCard.js
```diff
@@ -63,20 +65,19 @@ function ResourceCard({
 
   // Sync IDs with stylesheet
   // eslint-disable-next-line react/prop-types
-  const VotingBlock = ({ id }) => {
-    const onUpvote = () => handleVote('upvotes', setUpVotes);
-    const downUpvote = () => handleVote('downvotes', setDownVotes);
+  const VotingBlock = ({ blockID }) => {
+    const onVote = voteDirection => handleVote(voteDirection, id, setUpVotes, setDownVotes);
```
kylemh, line 69: I still actually prefer the separate methods so that we're not defining anonymous methods that don't compile down (see line 80 and 105).

Feel free to keep `onVote`, but I still think defining:

```js
const onUpvote = onVote('upvote');
const onDownvote = onVote('downvote');
```

is best practice regarding the mantra of always naming your methods.

kylemh (commented):
I feel like this is great, but I would love to see some integration tests regarding upvoting and downvoting both as a logged-in user and as a guest (assert that the login modal renders).

Eager to hear what Aaron has to say about the auth stuff.

components/Cards/ResourceCard/ResourceCard.js
chynh: okay no problem. I'll make separate functions for each.

common/constants/api.js
chynh: I am no expert in REST architecture but I guess `PUT` is more suitable since we are modifying a resource? But more importantly, this endpoint is configured to accept only `PUT` method AFAIK. I could be wrong so Aaron can verify this. Using other types of methods would yield `405` response.

chynh:
> I feel like this is great, but I would love to see some integration tests regarding upvoting and downvoting both as a logged-in user and as a guest (assert that the login modal renders).
> 
> Eager to hear what Aaron has to say about the auth stuff.

Agree. I can write integration tests once auth issue is resolved.

common/constants/api.js
kylemh: That makes sense to me. In my mind we'd be modifying a table of votes corresponding to a record, but if we're actually modifying a column in a record on the table for resources, PUT is totally the right call.

aaron-suarez: Yeah, the idea is that it's modifying the vote count on the resource. In the actual data model, a vote is its own thing because you need to distinguish whether a certain user has already voted. But the API doesn't reveal who has voted for what, it only reveals the number of votes. So we went with PUT because as far as the API is concerned, you're modifying a resource, not creating/deleting a vote

aaron-suarez:
@kylemh Re: the auth issue. I thought that we had it set up to where the front end would send your cookie of you being logged in alongside the request? If that's not happening, it's a front end issue. If it is sending your cookie and the API isn't accepting it, I'll have to look into that.

@chynh Can you please describe in more detail about why the vote functionality doesn't work when you're logged in unless you include your API key?

aaron-suarez:
Also, we should get the modal prompting you to log in done first before this is merged (or roll that work into this PR, whichever makes sense). To be fair, I haven't pulled this down and ran it, so if that's already done, don't mind me. If it's not done, I believe it should be

common/constants/api.js
kylemh: It comes along for the ride via a client-side set authorization head: `Authorization: `Bearer ${token}``.

Screen Shot 2020-10-09 at 1 31 47 PM

kylemh:
It comes along for the ride via a client-side set authorization head: `Authorization: `Bearer ${token}``.

Screen Shot 2020-10-09 at 1 31 47 PM

aaron-suarez:
Ok, so it's sending the right authorization header but it's still not letting you upvote/downvote without an API key?

chynh:
> Ok, so it's sending the right authorization header but it's still not letting you upvote/downvote without an API key?

Correct. As you can see, the authorization header is properly set.

> Also, we should get the modal prompting you to log in done first before this is merged (or roll that work into this PR, whichever makes sense). To be fair, I haven't pulled this down and ran it, so if that's already done, don't mind me. If it's not done, I believe it should be
That part is already taken care of. If you are not logged in, modal would pop up, prompting you to log in first.

aaron-suarez:
>If you are not logged in, modal would pop up, prompting you to log in first.

Cool, I just tried it, it looks great!

Let me see if I can figure out what's going on with the authorization issue

aaron-suarez:
Ah ha, @chynh you are absolutely right! I see this in the logs
```
sqlalchemy.exc.IntegrityError: (psycopg2.errors.NotNullViolation) null value in column "voter_apikey" violates not-null constraint
DETAIL:  Failing row contains (null, 21, upvotes).
```
Looks like there's a slight bug when you are trying to upvote a resource with only an `authorization` cookie, but not an API key. I'll get that squared away for you now if it's not terribly complicated to fix.

aaron-suarez:
I created a unit test for upvoting and downvoting a resource using the authorization header, and it fails in the same way. See OperationCode/resources_api#394

I'm going to finish that PR and get it deployed ASAP

kylemh:
#### 📝Changed routes:


| [**/resources/\[page\]**](https://storybook-git-enable-voting-resources.operation-code.vercel.app/resources/[page]) |
|:-:|
|<a href="https://storybook-git-enable-voting-resources.operation-code.vercel.app/resources/[page]">Screenshot of /resources/[page]</a><br /><sup><a href="https://images.weserv.nl?url=https://s.microlink.io/NL3QZJCPFD_rHKnMFYrJvkdcahg_.png">(view full size)</a> |







Commit 67af3a52485d91d1f511827a62a891b5605b0345 (https://storybook-git-enable-voting-resources.operation-code.vercel.app).

chynh:
@aaron-suarez Still getting 401 responses. did some investigation on my own and looks like the issue is from `authenticate` decorator on this endpoint, more specifically when you call `jwt_to_key`. I was seeing `InvalidSignatureError` get raised.
https://github.com/OperationCode/resources_api/blob/main/app/api/auth.py#L94-L113

aaron-suarez:
@chynh I just deployed the authorization fix to production, I hadn't done that yet. Let me know if you're still getting errors.

chynh:
> @chynh I just deployed the authorization fix to production, I hadn't done that yet. Let me know if you're still getting errors.

Unfortunately, I am still getting an error.

aaron-suarez:
When we get a chance it would be nice to jump on a call and do a screenshare because it worked fine for me just now. I'm on an airplane at the moment so now is not a good time, but I'll help you out as best I can when I can. 

Feel free to chat with me on slack about it if we don't want to blow up this PR with comments trying to troubleshoot it

chynh:
> When we get a chance it would be nice to jump on a call and do a screenshare because it worked fine for me just now. I'm on an airplane at the moment so now is not a good time, but I'll help you out as best I can when I can. 
> 
> 
> 
> Feel free to chat with me on slack about it if we don't want to blow up this PR with comments trying to troubleshoot it

You mean you were able to upvote/downvote with authorization header from dev frontend?

Without access to the server log, I don't have much further insight into what could be causing the issue.

I'll try to find time tomorrow to reach out to you on slack. In the meantime, I would appreciate if you could look at the log again to see if anything stands out.

aaron-suarez:
Does the "dev frontend" authenticate to production backend (i.e., https://api.operationcode.org/auth/login)? That's the cookie you'll need to communicate with the resources API

chynh:
> Does the "dev frontend" authenticate to production backend (i.e., https://api.operationcode.org/auth/login)? That's the cookie you'll need to communicate with the resources API

Yep. that was issue. Frontend is configured to make calls to `api.staging.operationcode.org` in dev. I can confirm that I was able to successfully upvote/downvote when authenticated via prod backend.

Although I haven't touched the current config since I don't fully understand the impact of changing it. Also added some integration test which I am not too sure about so feel free to provide any suggestions as well.

kylemh:
@aaron-suarez we'll need staging to work as well tho

aaron-suarez:
@kylemh staging works with staging and prod works with prod. If you get your token from `api.staging.operationcode.org/auth/login` then that token works with `resources.staging.operationcode.org/api/v1/resources`. If you get your token from `api.operationcode.org/auth/login` it works with `resources.operationcode.org/api/v1/resources`. I'm pretty sure that's the case anyway. If it's not, let me know.

From what I was reading, it looked like the problem is that somewhere the front end is trying to use the staging backend API to log in but then use the prod resources API. I could try to find a way to where either token works with either of the resources APIs but I'm not sure I understand the benefit

kylemh:
oh oopsie

https://github.com/OperationCode/front-end/pull/1286

chynh:
@kylemh do you know what is going on with Cypress? Not sure why it's failing towards the end of resources test suite. I can confirm the new tests work when I ran it in dev.

kylemh:
I think it's an issue with data. I haven't looked into the test much deeper, but my guess is we're asserting something regarding a specific resource which will always be flakey.

components/Form/Select/Select.js
```diff
@@ -93,7 +93,7 @@ export default function Select({
    * @returns {string[]}
    */
   const getValueFromMulti = () => {
-    return options.filter(option => fieldValue.includes(option.value));
+    return options.filter(option => (fieldValue || []).includes(option.value));
```
chynh, line 96: Resetting search params would basically causing this error.

Screen Shot 2020-10-12 at 5 22 36 PM

cypress/integration/resources.spec.js
```diff
@@ -30,9 +35,11 @@ describe('resources', () => {
   const assertThatListHasSameItems = () => compareResourceNames({ shouldBeEqual: true });
 
   beforeEach(() => {
+    cy.wait(2000); // eslint-disable-line cypress/no-unnecessary-waiting
```
chynh, line 38: I don't know if this ultimately solves anything but at least it looked like it allowed some failing tests to be more stable when testing in dev.

cypress/integration/resources.spec.js
```diff
@@ -30,9 +35,11 @@ describe('resources', () => {
   const assertThatListHasSameItems = () => compareResourceNames({ shouldBeEqual: true });
```
chynh, line 35: Looks like this function is failing consistently. Not sure why since `assertThatListHasDifferentItems` seems to be working fine.

kylemh:
I'm doing some dedicated OSS work tomorrow evening. I'll check it out then!

chynh:
Login modal style was a little off so I fixed it.
## Before
Screen Shot 2020-10-18 at 3 52 23 PM

## After
Screen Shot 2020-10-18 at 3 52 50 PM

cypress/integration/resources.spec.js
```diff
@@ -124,7 +131,7 @@ describe('resources', () => {
     });
 
     // Starting page
-    assertThatListHasSameItems();
+    // assertThatListHasSameItems();
```
chynh, line 134: I commented this out to see if the integration test passes. Seems `resources.spec.js` is now passing but saw that `join.spec.js` failed. Not sure why.

kylemh:
@chynh I've gotten integration tests working again on `main`. Thanks for the help @aaron-suarez

cypress/integration/resources.spec.js
```diff
@@ -239,4 +245,63 @@ describe('resources', () => {
...
+    cy.wait('@upvote');
+
+    cy.get('@upvote').then(({ status, response }) => {
+      expect(status).to.eq(200);
+      cy.get('@currentUpVoteCountText').should(
+        'not.eq',
+        `Number of Number of upvotes:${response.body.resource.upvotes}`,
+      );
+    });
+  });
```
kylemh, line 284: I feel like these are a bit mixed.

`> will prompt user to login when voting without being logged in` but then you assert that the login form works

I would consider getting rid of the login here, remove the preserved cookie, and simply make the act of logging in and then voting one test

kylemh (approved):
Optional feedback. Resolve via PR if you wish!

Amazing work, as always @chynh

PR #1334, opened 2020-10-16
Changed paid to free in resources page

### LINKED ISSUE #1284: Change `paid` to `free` in /resources
# Feature request

There is a pending change in the Resources API to change `paid` to `free` because it makes more sense as to what it means. See https://github.com/OperationCode/resources_api/pull/379 When this change gets deployed, it will break /resources. We'd like to avoid that happening.

## Proposed solution

This will be a two part solution. First, we need to change `pages/resources/[page].js` to send both `paid` and `free` in its requests so that way it will work with either. Once the API change happens, the front end should still be working. Then, we can go in and completely remove all references to `paid` on the front end. So those will be two different Pull Requests. The first will be deployed, then the Resources API will be deployed, then the second will be deployed.

## Additional context

The ultimate idea is to have a checkbox that says `free`, and that should map directly to the api response of whether you have to pay for a resource or not. If `free` is `true`, then you know that you don't need to pay to access that resource. If `free` is false, then you must pay to access it.

BhuwanChandra:
# Description of changes
Changed `paid` to `free` in resources page

# Issue Resolved

Fixes #1284 

## Screenshots/GIFs
NA


kylemh (commented):
Amazing work! You'll just need to change the reference to the correct Test ID in the resources Cypress test.

kylemh (approved):
Flawless

PR #1337, opened 2020-10-24
fix upvote/downvote integration test

chynh:
# Description of changes

Moved logging in and upvoting/downvoting into the same block as suggested in #1283 
small fix to the test that was incorrect

# Issue Resolved

Fixes #NA


cypress/integration/resources.spec.js
```diff
@@ -259,48 +260,38 @@ describe('resources', () => {
...
-
     cy.findAllByTestId(UPVOTE_BUTTON).first().click({ force: true });
     cy.wait('@upvote');
 
     cy.get('@upvote').then(({ status, response }) => {
       expect(status).to.eq(200);
       cy.get('@currentUpVoteCountText').should(
         'not.eq',
-        `Number of Number of upvotes:${response.body.resource.upvotes}`,
+        `YesNumber of upvotes:${response.body.resource.upvotes}`,
```
kylemh, line 277: I'm confused by this assertion and the one on line 294.

Why assert for content to not equal specific content after the API response returns. Shouldn't we know what the value is after an upvote/downvote and assert against that?

chynh: I see what you are saying. I did this because there is no clean way of grabbing the current vote count.

If we wrap the actual count with a `span` or something and give it a test id, I guess we can assert for the content to be equal a response. What do you think?

https://github.com/OperationCode/front-end/blob/ca6998364c0aef1ecfd1cb7191ac5a948cba390e/components/Cards/ResourceCard/ResourceCard.js#L99

chynh: Oh and also the each vote button would have to be clicked twice since depending on the vote status, the count could either increase/decrease by 1.

kylemh: We could wrap the number in a span and give it a testID to quickly grab the actual vote number.

Then we could findByTestId before and after the actual vote to ensure the count increments or decrements.

chynh: Aren't that asserting for the value to not equal? Maybe I misunderstood what you were saying but I thought we were aiming at asserting for the value to be equal.

On a side note, I just noticed that `userVote` prop is not being passed into ResourceCard. If we had that information, it would definitely make things easier.

kylemh: You can essentially store the value by a variable and assert that after the PUT, that the value becomes -1 or +1

chynh:
Hey @kylemh, do you have any futher comments on this PR? If not, could you merge this in? Tryna get that last PR before the end of Hackberfest. Thanks!

kylemh (approved):

cypress/integration/resources.spec.js
```diff
@@ -261,38 +263,33 @@ describe('resources', () => {
...
 
     cy.get('@upvote').then(({ status, response }) => {
       expect(status).to.eq(200);
-      cy.get('@currentUpVoteCountText').should(
-        'not.eq',
-        `YesNumber of upvotes:${response.body.resource.upvotes}`,
-      );
+      cy.get('@currentUpVoteCountText').then(upVoteText => {
+        const upVoteNumber = parseInt(upVoteText, 10);
+        expect(response.body.resource.upvotes).to.be.oneOf([upVoteNumber + 1, upVoteNumber - 1]);
```
kylemh, line 275: wait why would it be one below or one above?

chynh: That's because we only allow 1 vote per user. So for example, if the user has already upvoted on a resource and clicks upvote again, the vote count would decrement by one whereas if the user is yet to vote or has downvoted on that resource, upvote count would increment by one.

I am working on a PR on resource api side to include user's current vote status when fetching resources. If that get's done, we can modify this test to either assert for current vote count + 1 or -1 depending on the current vote direction.

kylemh: Sounds good.

PR #1356, opened 2020-11-08
display user vote direction

chynh:
# Description of changes

* Display authenticated user's vote information 
* Modify integration test/unit test related to the change

# Issue Resolved

Fixes #NA

## Screenshots/GIFs

vote_information

### DIFF
--- a/components/Cards/ResourceCard/ResourceCard.js
+++ b/components/Cards/ResourceCard/ResourceCard.js
@@ -62,15 +62,18 @@ function ResourceCard({
 }) {
   const [upVotes, setUpVotes] = useState(upvotes);
   const [downVotes, setDownVotes] = useState(downvotes);
-  const didUpvote = userVote === possibleUserVotes.upvote;
-  const didDownvote = userVote === possibleUserVotes.downvote;
+  const [userVoteDirection, setUserVoteDirection] = useState(userVote);
+
+  const didUpvote = userVoteDirection === possibleUserVotes.upvote;
+  const didDownvote = userVoteDirection === possibleUserVotes.downvote;
 
   const DESKTOP_VOTING_BLOCK = 'desktopVotingBlock';
 
   // Sync IDs with stylesheet
   // eslint-disable-next-line react/prop-types
   const VotingBlock = ({ blockID, resourceID }) => {
-    const onVote = voteDirection => handleVote(voteDirection, resourceID, setUpVotes, setDownVotes);
+    const onVote = voteDirection =>
+      handleVote(voteDirection, resourceID, setUpVotes, setDownVotes, setUserVoteDirection);
     const onUpvote = () => onVote('upvote');
     const onDownvote = () => onVote('downvote');
 
--- a/components/Cards/ResourceCard/ResourceCard.js
+++ b/components/Cards/ResourceCard/ResourceCard.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 import { string, number, func, oneOf, oneOfType, array, bool } from 'prop-types';
 import classNames from 'classnames';
 import Accordion from 'components/Accordion/Accordion';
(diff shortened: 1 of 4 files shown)


components/Cards/ResourceCard/ResourceCard.js
```diff
@@ -62,15 +62,18 @@ function ResourceCard({
 }) {
   const [upVotes, setUpVotes] = useState(upvotes);
   const [downVotes, setDownVotes] = useState(downvotes);
-  const didUpvote = userVote === possibleUserVotes.upvote;
-  const didDownvote = userVote === possibleUserVotes.downvote;
+  const [userVoteDirection, setUserVoteDirection] = useState(userVote);
```
kylemh, line 65: This line shouldn't be necessary. Keep in mind, `userVote` is a prop because it's not just state for the current session, but their vote direction also needs to be provided by the back-end. In other words, if you upvote a resource, close your browser, re-open the browser, and go back to the same resource - it should stills how that resource as upvoted.

The state for a ResourceCard's userVote should be managed via the resource page file for that reason and it's initial state should be derived from the request for data on the resources.

cypress/integration/resources.spec.js
```diff
@@ -265,14 +265,20 @@ describe('resources', () => {
 
     cy.findAllByTestId(UPVOTE_COUNT).first().invoke('text').as('currentUpVoteCountText');
 
-    cy.findAllByTestId(UPVOTE_BUTTON).first().click({ force: true });
-    cy.wait('@upvote');
+    cy.findAllByTestId(UPVOTE_BUTTON).first().as('upvoteButton');
```
kylemh, line 268: I think you can simplify things by:

1) Making two separate tests for upvote and downvote.
2) Add a different testid if the button is active
3) Check that the different testid renders after click and that the vote count went up
4) Presuming you follow my critique in the other comment, you can ditch the test regarding the actual response and instead just reload the page after a vote to assert the vote still exists. This works out because the initial state will derive from the response and a reload ensures it's not the previous client state tick.

chynh: Just so we are on the same page before I jump ahead:
1. That would require persisting cookie as I did before, correct?

3. So This is a little tricky provided that the test strategy is staying the same (grab the first resource, upvote first, then downvote). The issue is the initial upvoting part since there is no guarantee of the initial vote direction so there should be some conditional logic by checking whether the button is active then to assert for current upvote count +1 or -1. 

Can you write a conditional logic by checking whether certain testid renders? I am not too familiar with having conditional flow in Cypress test.

So it just dawned on me that I could utilize the response that I got from API call. Say I clicked upvote and in the response that I got,  if `current_vote_direction` is `upvote` that means the upvote count will increment by 1 whereas if it is `null`, the upvote count will decrement by 1. 

Let me know what you think.

kylemh: 1) No, you can use `cy.login()` per test.

2) I don't mean to add the conditional logic in Cypress, but in the actual source code. You can totally add whatever logic you need, to any source code file, surrounding `data-testid`. Use it to your advantage to make the test code as simple as possible.

kylemh:
#### 📝Changed routes:


| [**/resources/\[page\]**](https://operation-code-git-include-vote-direction.operation-code.now.sh/resources/[page]) |
|:-:|
|<a href="https://operation-code-git-include-vote-direction.operation-code.now.sh/resources/[page]">Screenshot of /resources/[page]</a><br /><sup><a href="https://images.weserv.nl?url=https://s.microlink.io/VCR4KW9EfmBRyQDKylt4NQZa9a4_.png">(view full size)</a> |







Commit 90b96d04b36a30570afdc5f972eebd1e5d77ca73 (https://operation-code-git-include-vote-direction.operation-code.now.sh).

cypress/integration/resources.spec.js
chynh: So if I understood this correctly, we would have data-testid `UPVOTE_BUTTON` or `UPVOTE_BUTTON_ACTIVE`(something to that effect) depending on the button's active status.

Going back to line 268, let's say we leave it as is and the button was clicked. Now that the button should have `data-testid=UPVOTE_BUTTON_ACTIVE`. How can I grab the same button since there is no guarantee that ` cy.findAllByTestId(UPVOTE_BUTTON_ACTIVE).first()` is going to target the same button?

cypress/integration/resources.spec.js
```diff
@@ -262,31 +261,44 @@ describe('resources', () => {
     cy.wait('@postLogin').its('status').should('eq', 200);
 
     cy.findByTestId(LOGIN_FORM).should('not.exist');
+  });
 
-    cy.findAllByTestId(UPVOTE_COUNT).first().invoke('text').as('currentUpVoteCountText');
+  it('will not allow user to upvote once logged in', () => {
```
kylemh, line 266: wait they can't upvote once they're logged in?

kylemh:
We killed the resources feature.

PR #1381, opened 2021-01-31
Create resource form

### LINKED ISSUE #1198: Resources page: form to upload a new resource
# Feature request

The API allows an authenticated user to upload a new resource or edit an existing one. This functionality should be available on the /resources page as well.

## Proposed solution

There should be a button that says "Add a new resource" or perhaps just "New Resource". When you click the button, if you're not logged in, you should get the same modal described in #1197. If you are logged in, you should see a form that allows you to upload a new resource. This can either be a dedicated page or an additional modal. We have a design for this floating around somewhere, and I'm hoping that makes it into this issue but I don't have it readily available right this moment.

I don't have an alternative solution to a basic form to fill in the information. If you think of one, feel free to suggest it, but the form seems like the most natural solution.

## Additional context

I'm not sure if the edit resource button/form should go in this same issue or if that should be a dedicated issue. Please let me know if I need to create another one or describe above, in more detail, what that will look like.

kylemh:
# Description of changes
Give users the ability to create resources

# Issue Resolved
Fixes #1198


kylemh:
#### 📝Changed routes:


| [**/resources/\[page\]**](https://operation-code-pq31141o7.vercel.app/resources/[page]) | [**/resources/create**](https://operation-code-pq31141o7.vercel.app/resources/create) |
|:-:|:-:|
|<a href="https://operation-code-pq31141o7.vercel.app/resources/[page]">Screenshot of /resources/[page]</a><br /><sup><a href="https://images.weserv.nl?url=https://s.microlink.io/XMs8C3D5uvJCm13lVZiAuzAuito_.png">(view full size)</a> |<a href="https://operation-code-pq31141o7.vercel.app/resources/create">Screenshot of /resources/create</a><br /><sup><a href="https://images.weserv.nl?url=https://s.microlink.io/fUu78Fldi09ceNonTSrtH1_5kMk_.png">(view full size)</a> |





1 deleted route:
- **/resources**

Commit 20ce0ec0ed6fb07f81982fcdb89b0b4e72ce9823 (https://operation-code-pq31141o7.vercel.app).

kylemh:
Hey @aaron-suarez 

I'm getting 404 responses when attempting to create resources as a logged in user. Feel free to lemme know if you see anything wrong with the code or request payload when playing with the preview deployment!

aaron-suarez:
I know this wasn't part of this PR, but I noticed when playing around with it that there's no link to join OC on this page and there ought to be I think.
Screen Shot 2021-01-31 at 4 43 20 PM

components/Forms/CreateResourceForm/CreateResourceForm.js
```diff
@@ -0,0 +1,149 @@
...
+          />
+
+          <Button
+            className={styles.submitButton}
+            type="submit"
+            theme="secondary"
+            disabled={isSubmitting}
+          >
+            Submit
+          </Button>
```
aaron-suarez, line 142: In addition to the submit button, we should also have a "back" or "X" or "cancel" button, in case the user decides against submitting their resource

aaron-suarez:
In the original issue I had said "I'm not sure if the edit resource button/form should go in this same issue or if that should be a dedicated issue"

Do we want to add a button to "Edit Resource" in this PR or open a separate issue for that?

components/Forms/CreateResourceForm/CreateResourceForm.js
```diff
@@ -0,0 +1,149 @@
...
+const free = { value: 'true', label: 'Free' };
+const paid = { value: 'false', label: 'Paid' };
+
+const costOptions = [free, paid];
+
+const validationSchema = Yup.object().shape({
+  name: Yup.string().required(validationErrorMessages.required).default(''),
+  url: Yup.string().required(validationErrorMessages.required).default(''),
+  category: Yup.string().required(validationErrorMessages.required).default(''),
+  languages: Yup.array().default(['']),
```
aaron-suarez, line 25: This doesn't matter for creating, but when we do a `PUT` to update a resource, if the languages are not meant to change, the client should omit the `languages` rather than sending `[]` because the latter will clear out all existing languages.

And also, we should probably remove the quotes here, because it should be a default of `[]` rather than `['']`

components/Forms/CreateResourceForm/CreateResourceForm.js
```diff
@@ -0,0 +1,149 @@
...
+          />
+
+          <Field
+            name="languages"
+            label="Programming language(s)"
+            placeholder="What language or languages are involved?"
+            options={languageOptions}
+            component={Select}
+            isDisabled={isSubmitting}
+            required
```
aaron-suarez, line 112: `languages` aren't relevant for every resource, so it's not required. I noticed you didn't put a `*` by the label, but won't this still make it fail validation if it's empty? Or does the default of `['']` cover that case and it's never empty?

pages/resources/[page].js
```diff
@@ -48,28 +52,36 @@ function Resources() {
...
 
-  const handleVote = (voteDirection, id, setUpVotes, setDownVotes) => {
+  const onVote = (voteDirection, id, setUpVotes, setDownVotes) => {
     setErrorMessage(null);
-    if (!hasValidAuthToken()) {
-      setIsModalOpen(true);
-      return;
+
+    if (!isLoggedIn) {
+      openLoginModal();
```
Comment (aaron-suarez) [inline pages/resources/[page].js:74]: When I click a `thumbs up` or `thumbs down` on a resource, it doesn't prompt me to log in.. which I thought it would do based on this code. Is that working for you?

components/Forms/CreateResourceForm/CreateResourceForm.js
kylemh: So this spot is for compatibility with the multi select component; however, in the submit handler we can translate your needs there much how we've done with cost --> free.

components/Forms/CreateResourceForm/CreateResourceForm.js
kylemh: Ah shoot. This was just a copy paste mistake. It was clearly not required in the screenshots

pages/resources/[page].js
Comment (kylemh) [reply pages/resources/[page].js:74]: ... hmmm it was for me. Maybe I had a regression I didn't notice.

Comment (kylemh) [reply pages/resources/[page].js:74]: Does it work for u on prod?

aaron-suarez:
I was clicking on the languages component trying to select multiple values. I don't think the multiselect is quite there yet. Then it showed me this error, which I'm not sure if that means anything to you?
Screen Shot 2021-01-31 at 5 25 07 PM

pages/resources/[page].js
Comment (aaron-suarez) [reply pages/resources/[page].js:74]: It works for me on prod... I'm not sure, maybe it's specific to vercel or something?

Comment (kylemh) [reply pages/resources/[page].js:74]: No, I've messed something up on this branch I mean since integration tests are failing. Just didnt realize during local dev

Comment (aaron-suarez) [reply pages/resources/[page].js:74]: No worries, happy to QA it and find those bugs 😄

components/Forms/CreateResourceForm/CreateResourceForm.js
```diff
@@ -0,0 +1,172 @@
...
+      actions.setSubmitting(false);
+
+      const resourcesAPIErrors = get(error, 'response.data.errors');
+
+      if (Array.isArray(resourcesAPIErrors)) {
+        const [firstError] = resourcesAPIErrors;
+
+        if (firstError['invalid-params'] && firstError['invalid-params'].params.includes('url')) {
+          onFailure('A resource with this URL already exists in our database!');
+        }
```
kylemh, line 82: It may be worth some energy thinking about the best way to handle this.

We don't do field-level server-side error messages, so I'll just be picking the first error I see in the response and trying to parse it.

Otherwise we'll give a generic: "Try again later 🤷‍♂️"

aaron-suarez: The error messages returned from the API are all created with the user in mind, I don't think there's anything wrong with just passing that error through to the user rather than hardcoding one on the front end.

To answer your question though, I think that there will only be one error if there's only one resource submitted. Until we support submitting multiple resources with one request, we can kick that can down the road I would think. And we may never support uploading multiple resources in one request on the front end, that would be ok too.

kylemh: The only problem I have with simply passing through the error is that it's not really language an end-user should be seeing:

```The following params were invalid: url. Resource id 2995 already has this URL.```

I'd rather see:

```A resource with this URL already exists in our database.```

aaron-suarez: Ok, fair enough. We need to keep that ID in there for programmers to see when things go wrong as they're developing on top of the API, so I wouldn't make that change on the backend.

Feel free to leave it like it is 😄

aaron-suarez:
@kylemh and I have talked about outstanding items and we're going to address them in a followup PR:

1. We're still going to add a join link to the login modal
2. We're going to add tests around this
3. We're maybe going to be adding a cancel button before you're finished submitting a resource, but we'll see
4. We've talked about possibly adding end-user error messages to be returned from the API and passed along directly to the end user. Again, we'll see. For now, the one error message is probably fine.

PR #1519, opened 2021-10-20
Refactor press links

### LINKED ISSUE #1511: Refactor "In The News" Section
[At this location](https://operationcode.org/press#in-the-news-link), we list articles where Operation Code has been mentioned in the media. We had an issue where some groupings had way more articles than other groupings, so the UI could get messy (one column is short and the other is very long):

Screen Shot 2021-10-16 at 6 31 38 PM

We tried assuaging this issue by creating a button that would collapse or expand content. This makes things like better, but it still looks ugly if everything is expanded. It also sucks to expand content and have the collapse button move away from you.

I'd like to adjust this UI to use [Radix UI's Tabs](https://www.radix-ui.com/docs/primitives/components/tabs) component where the tab buttons are the groupings (NYC, Boston, etc.) and the list is simply fully expanded. This will give the user total control of what grouping they care to see, and provide them the ability to quickly go to another list if they are shocked by the length. It also will scale better if we add another grouping or two.

I don't have designs, so use your best judgement.

xevenheaven:
# Description of changes
* Use react-tabs for press articles
* Remove `ArticleGroup` as it's no longer used

# Issue Resolved
Fixes #1511

## Screenshots/GIFs
PressLinks

### DIFF
--- a/components/Press/PressLinks/PressLinks.module.css
+++ b/components/Press/PressLinks/PressLinks.module.css
@@ -59,3 +59,27 @@
 .logos li::marker {
   display: none;
 }
+
+@media screen and (--medium-viewport) {
+  .logos .flexContainer {
+    width: 60ch;
+  }
+}
+
+@media screen and (--small-viewport) {
+  .logos {
+    width: 80%;
+  }
+
+  .logos .flexContainer {
+    width: 100%;
+  }
+
+  .logos .tabsList {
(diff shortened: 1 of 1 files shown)


kylemh:
image

Whoops. I had not yet checked mobile. What do you want to do about this? Feel free to act on whatever thing is best.

PR #1520, opened 2021-10-20
Fix broken LogRocket image on readme

xevenheaven:
# Description of changes
The current LogRocket image is broken on the readme, hence updating to the newly added svg file instead.

# Issue Resolved
Fixes #NA

## Screenshots/GIFs
NA


xevenheaven:
Hi @kylemh, as the svg is not yet uploaded to the Github server, it will still look broken on this PR.

Once this PR is merged into main, the svg will then be available at this url (https://raw.githubusercontent.com/OperationCode/front-end/main/public/static/images/logrocker.svg) and it should render correctly, similar to how the Chromatic and Sentry logos are being used. 🙂

PR #1528, opened 2021-10-29
Updated CSS :focus to :focus-visible

### LINKED ISSUE #1514: Change focus styles to be applied on focus-visible instead
Focus rings should only be seen by keyboard navigating users! To make it so, we should add styles to  the `:focus-visible` pseudo-selector instead of `:focus`.

Unfortunately, Safari _**STILL**_ doesn't supports `:focus-visible` 😭. This means we need to add the feature with [backwards compatibility](https://www.tpgi.com/focus-visible-and-backwards-compatibility/) OR [polyfill](https://github.com/WICG/focus-visible) the feature. 

Polyfilling _correctly_ is complicated and potentially costly for ALL users (even ones that have no need to download that polyfill because they use a browser that supports the feature). So, let's go the backwards compatible route instead.

MartinL551:
# Description of changes

# Issue Resolved

Fixes #1514

## Screenshots/GIFs

image

### DIFF
--- a/components/Accordion/Accordion.module.css
+++ b/components/Accordion/Accordion.module.css
@@ -69,7 +69,7 @@
   text-decoration: underline;
 }
 
-.accordionToggleButton:focus {
+.accordionToggleButton:focus-visible {
   border: 1px solid var(--primary);
   border-radius: inherit;
   outline: none;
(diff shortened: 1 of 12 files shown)


kylemh (approved):
I don't think there are any issues with this PR, but integration tests are failing. I'll have to confer with other OC staff on the state of infrastructure!

PR #1529, opened 2021-10-30
Fix code schools search form collisions

### LINKED ISSUE #1527: Code Schools page has a small style bug
# Bug Report
When visiting operationcode.org/code_schools and opening the US State dropdown, it can collide with badges on code school cards.

The dropdown should always cover any UI below it.

### Screenshots
Screen Shot 2021-10-24 at 10 51 20 PM

dlim87:
# Description of changes
This should hopefully fix the issue where the ribbons on the schools in the code_schools page were still being shown on top of an active search form. Unfortunately, I couldn't figure out how to use Apiary to mock the back-end, and allow the page to be rendered, so I'm flying a bit blind here, but I'm fairly confident that this should fix it. The ribbons only have a z-index of 1, so giving the searchbox a z-index of 2 should supercede it

# Issue Resolved
Fixes #1527 

# Pictures
image

### DIFF
--- a/common/styles/variables.css
+++ b/common/styles/variables.css
@@ -35,5 +35,5 @@
   --borderRadius: 3px;
   --navBarHeight: 65px;
   --navBarTopGutterHeight: 20px;
-  --typicalMaxWidth: 700px;
+  --typicalMaxWidth: 70ch;
 }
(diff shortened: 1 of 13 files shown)


kylemh:
@dlim87 the issue around Apiary was that data on staging had no schools with the ribbons? I'll fix that.

dlim87:
> @dlim87 the issue around Apiary was that data on staging had no schools with the ribbons? I'll fix that.

I think the issue was that something was down. It seemed like it was also causing the integration specs to fail, but fixed itself the next day

PR #1530, opened 2021-10-31
Change mas-width to use ch instead of pixels

### LINKED ISSUE #1513: Change max-width large pixel amounts to 70ch where relevant
There are many elements that have large pixel amounts as values for a max-width property to make the content seem bordered, consistent, and collected. In many of those places, a value of `70ch` is better than `XXXpx`. For example, in `globalStyles.css` the max-width of the `p` tag is responsively set to different values with media queries. Instead, it should simply always be set to `max-width: 70ch`. There are other locations where max-width is used in the same way that can be changed.

dlim87:
# Description of changes
Changes the typicalMaxWidth css property to be 70ch, instead of 700 pixels, and change a few calls which had large pixel counts (above 500px) to use it. In practice, it appears that with the current font style/sizes, 70ch works out to about 773px

Changes:
* Contact page (Not the contact section on every page)
* Upgrade browser popup
* partner Logos

Existing calls for typicalMaxWidth:
* p tags (as mentioned in tickets)
* create resource 
* get involved

Large pixel count elements that were NOT changed:
* media-query breakpoints
* imageCard (this made them render as only 1 to a row at max width, which was not as good UI)
* Container.content (this broke the stories section)
* FooterWrapper (this broke alignment)
* desktopNavContainer (this removed all the white space from the navbar)

# Issue Resolved
Fixes #1513 

## Screenshots/GIFs
I'm not so sure about using this for partner logos, as it knocks the alignment off by a bit:
Screenshot 2021-10-30 203033
<details><summary>Click here for other pictures (collapsed for size)</summary>
Screenshot 2021-10-30 202327

Screenshot 2021-10-30 202324

Screenshot 2021-10-30 202239

Screenshot 2021-10-30 205151

Screenshot 2021-10-30 202727
</details>

### DIFF
--- a/common/styles/variables.css
+++ b/common/styles/variables.css
@@ -35,5 +35,5 @@
   --borderRadius: 3px;
   --navBarHeight: 65px;
   --navBarTopGutterHeight: 20px;
-  --typicalMaxWidth: 700px;
+  --typicalMaxWidth: 70ch;
 }
--- a/components/UpgradeBrowserOverlay/UpgradeBrowserOverlay.module.css
+++ b/components/UpgradeBrowserOverlay/UpgradeBrowserOverlay.module.css
@@ -11,7 +11,7 @@
   text-align: center;
   top: 50%;
   transform: translate(-50%, -50%);
-  width: 750px;
+  width: var(--typicalMaxWidth);
 }
 
 .overlay {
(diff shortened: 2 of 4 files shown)


kylemh (approved):

kylemh:
@dlim87 I'm really sorry I didn't get to these before Hacktoberfest ended... I've been very busy the last few days 😞

PR #1541, opened 2021-11-17
Run with React 18

eps1lon:
I don't recommend merging this PR right now. You should carefully audit compatibility of each library used. And of course wait for React 18 to become stable.
For example, `@reach/router` is using `componentWillReceiveProps` which is deprecated generally and more likely to cause actual problems with concurrent rendering.

# Description of changes
Uses React 18 and fixes failing tests.

The first commit should be merged separately. It prevents leaking of tasks between tests when a snapshot test is performed.

https://github.com/OperationCode/front-end/pull/1541#discussion_r751079118

# Issue Resolved
https://twitter.com/kylemh_/status/1460717366116073476

## Screenshots/GIFs

# What's missing

### login

For example, https://github.com/OperationCode/front-end/blob/e0e6dc0b4e33ec1b6bda697c9d4529bb81c71c44/components/Nav/Nav.js#L45 causes an hydration mismatch because the server renders the UI for non-logged in users but the client renders the logged-in view. This was causing hydration mismatches in React 17 as well though.

For React 18 you need to find a way to ignore these console errors and the dev overlay added by next (since the dev overlay acts as a modal).


components/Forms/RegistrationForm/__tests__/RegistrationForm.test.js
```diff
@@ -52,13 +52,19 @@ describe('RegistrationForm', () => {
...
+    const submit = getByText('Submit');
+    fireEvent.click(submit);
+
+    // formik switches to `isSubmitting` in a microtask.
+    // TODO: This should probably synchronously switch to being disabled.
+    await act(async () => {
+      await Promise.resolve();
+    });
 
-    const submit = await findByText('Submit');
```
eps1lon, line 61: The behavior of the component under test didn't actually changed. The tests only relied on the microtask queue being flushed via `await ...`. The `findByText` was actually irrelevant since the element was already in the DOM (see `getByText('submit')` from earlier.

We only need to flush microtasks to schedule the state update for `isSubmitting`. However, in React 17 those state updates where synchronous so we didn't need to do anything but assert on the outcome. In React 18 **every** state update must be wrapped in `act`. 

So the change can be more minimal with

```js
await act(async () => {
  await findByText('Submit');
});
```
but I opted to make the test more clearer for future reference.

components/Forms/RegistrationForm/RegistrationForm.js
```diff
@@ -214,7 +214,7 @@ function RegistrationForm({ initialValues, onSubmit, onSuccess }) {
             />
           </div>
 
-          {errorMessage && <Alert type="error">{errorMessage}</Alert>}
+          {Boolean(errorMessage) && <Alert type="error">{errorMessage}</Alert>}
```
eps1lon, line 217: Workaround for https://github.com/facebook/react/issues/22784

kylemh:
Sorted in #1650 thanks for trailblazing with us @eps1lon

PR #1552, opened 2021-12-07
Updating Accordion component to Radix Accordion

### LINKED ISSUE #1509: Replace Accordion Component with Radix UI's Accordion
We have and use an Accordion component in many places.

While we had the best intentions, there are minor accessibility issues with the Accordion. For example, the accordion header is listed before the button. This would mean people that use assistive technology would need to hear about the potential accordion, go to the button to open it, and back up to re-hear the title before being able to navigate into the actual content.

Rather than fix our component and implementation, I'd like to simply delete our component and re-implement a styled instance of [Radix UI's Accordion](https://www.radix-ui.com/docs/primitives/components/accordion) and update any stories and integration tests. Feel free to remove unit tests.

jakecs4:
# Description of changes
This PR updates the previous Accordion component to the Radix styled component as team members wanted this.

# Issue Resolved
Fixes #1509 

## Screenshots/GIFs
Screen Shot 2021-12-06 at 8 53 47 PM


components/Accordion/Accordion.js
```diff
@@ -45,47 +45,49 @@ function Accordion({ accessibilityId, className, content, hasAnimationOnHover })
...
+              [styles.hover]: hasAnimationOnHover,
+            })}
+            onClick={toggleAccordionContent}
+            aria-controls={contentId}
+            aria-expanded={isContentVisible}
+            data-testid={ACCORDION_TOGGLE_BUTTON}
+            id={accordionId}
+          >
+            <div className={styles.headingContainer}>{content.headingChildren}</div>
+            {isContentVisible ? (
```
kylemh, line 62: Radix exposes an HTML attribute on the trigger that will allow us to style the transition rather than maintain our own local state (`isContentVisible`). Check out their docs to see how they do it in their example, which is really similar to ours!

https://user-images.githubusercontent.com/9523719/145086636-afa9e65e-e565-474b-aea1-b3ea3b4e5803.mov

components/Accordion/Accordion.js
```diff
@@ -45,47 +45,49 @@ function Accordion({ accessibilityId, className, content, hasAnimationOnHover })
...
+        <RadixAccordion.Header>
+          <RadixAccordion.Trigger
+            className={classNames(styles.headingContainer, {
+              [styles.hover]: hasAnimationOnHover,
+            })}
+            onClick={toggleAccordionContent}
+            aria-controls={contentId}
+            aria-expanded={isContentVisible}
+            data-testid={ACCORDION_TOGGLE_BUTTON}
+            id={accordionId}
```
kylemh, line 59: Of all these attributes, only `data-testid` is necessary (for testing purposes).

We should be able to remove the others, and that was the point of wanting to use the Radix's accordion. Let them do all the logic. We just need to style it.

components/Accordion/Accordion.js
```diff
@@ -45,47 +45,49 @@ function Accordion({ accessibilityId, className, content, hasAnimationOnHover })
...
+        <RadixAccordion.Content>
+          <section
+            aria-hidden={!isContentVisible}
+            className={classNames(styles.accordionContent, {
+              [styles.visible]: isContentVisible,
+              [styles.invisible]: !isContentVisible,
+            })}
+            data-testid={ACCORDION_CONTENT}
+            id={contentId}
+            style={{ display: isContentVisible ? 'block' : 'none' }}
```
kylemh, line 84: You'll only want `className` and `data-testid` defined.

The value for `className` will also change quite a bit since you won't need to change visible/invisible styles. Radix will handle that for us.

components/Accordion/Accordion.module.css
```diff
@@ -13,6 +13,7 @@
 }
 
 .headingContainer {
+  all: unset;
```
kylemh, line 16: This _wont_ be what we want since we have some global CSS in use!

Additionally, the accordion trigger is missing `:focus-visible` styles and it doesn't have a pointer cursor on hover.

kylemh (commented):
A lot of good work here, but it looks like the linter is mad and there's a bit more work needed to let Radix be in charge.

components/Accordion/Accordion.js
jakecs4: Updated

components/Accordion/Accordion.js
jakecs4: Updated

components/Accordion/Accordion.js
jakecs4: Updated

components/Accordion/Accordion.module.css
jakecs4: Updated

components/Accordion/Accordion.js
```diff
@@ -32,60 +28,72 @@ Accordion.defaultProps = {
...
-        style={{ display: isContentVisible ? 'block' : 'none' }}
-      >
-        {content.bodyChildren}
-      </section>
-    </Card>
+      <RadixAccordion.Item value={accordionId}>
+        <RadixAccordion.Header>
+          <RadixAccordion.Trigger
+            className={classNames(styles.headingContainer, {
+              [styles.hover]: hasAnimationOnHover,
```
kylemh, line 78: Since you're no longer using our Card component, you dont need to expose `hasAnimationOnHover` as a prop to our accordion component, and you can simply always apply hover styles by changing the CSS to be:

```css
.headingContainer {
  /* styles */
}

.headingContainer:hover {
  cursor: pointer;
}
```

and then adjusting the className property value to be simply `styles.headingContainer`

kylemh (commented):
When sharing the Radix docs I didn't mean for you to do an _exact_ copy-paste including the other tooling they use. I just wanted to share how you can compose the underlying elements and avoid extra attributes.

You should be able to accomplish the needed stylings with our existing styling solution and icon. So, you shouldn't use or install `@radix-ui/react-icons` nor `@stitches/react`.

You added focus stylings, but the styles themselves are not correct. You should compare what you have on your local machine to what is currently on https://operationcode.org/faq

You're getting closer!

kylemh (commented):
Last thing is needing to uninstall the unused dependencies.

kylemh (approved):

kylemh (approved):

PR #1592, opened 2022-04-10
Replace all instances of react-modal with Radix UI's Dialog

### LINKED ISSUE #1510: Replace Modal Component with Radix UI's Dialog
We have and use a Modal component in a few places.

No real issues with `react-modal` that I'm aware of, but I'd rather use Radix for most/all primitives.

This ticket is to replace our Modal component and re-implement a styled instance of [Radix UI's Dialog](https://www.radix-ui.com/docs/primitives/components/dialog) and update any stories and integration tests. Feel free to remove unit tests.

florahon:
# Description of changes
Updates the shared Modal component to use Radix UI's Dialog component and removes all instances & imports of react-modal. Also updated corresponding unit tests and Cypress CI tests for modal component and instances (UpgradeBrowserOverlay, Resources page, and Code Schools page). 

**Note**: Currently running into an issue where I'm passing all Cypress tests locally, but the "join" tests have been failing on the GitHub checks. 

# Issue Resolved

Fixes #1510

## Screenshots/GIFs
Examples of replaced modal component with same styling, on resources and code schools pages:
Screen Shot 2022-04-10 at 11 25 02 AM
Screen Shot 2022-04-10 at 11 25 26 AM


cypress/integration/code_schools.spec.js
```diff
@@ -70,6 +71,9 @@ describe('code schools', () => {
       cy.findAllByTestId(SCHOOL_CARD_LOCATION_LIST_MODAL_BUTTON).first().click();
       cy.findByTestId(MODAL_CONTENT).should('exist').and('be.visible');
       cy.findAllByTestId(SCHOOL_LOCATION_LIST_ITEM).should('have.length.greaterThan', 1);
+      cy.findByTestId(MODAL_OVERLAY).click({ force: true });
```
kylemh, line 74: nice expansion of coverage here

cypress/integration/modal.spec.js
```diff
@@ -22,4 +23,11 @@ describe('when the server responds successfully', () => {
     cy.findByTestId(CLOSE_BUTTON).click();
     cy.findByTestId(MODAL_CONTENT).should('not.exist');
   });
+
+  it('closes the modal when the overlay is clicked', () => {
```
kylemh, line 27: Let's also add a test for clicking the close button.

kylemh: After adding that test, I think it'd be best to change the other tests to use the close button instead of the overlay. It's good to have regression coverage for clicking the overlay in the modal spec specifically, but force clicks in Cypress can be flakey. I think it'd be better to target the actual close button for other integration tests.

package.json
```diff
@@ -107,7 +107,7 @@
     "chromatic": "^6.0.4",
     "cross-env": "^7.0.3",
     "css-loader": "^6.4.0",
-    "cypress": "^8.6.0",
+    "cypress": "^9.5.3",
```
kylemh, line 110: nice! ty!

cypress/integration/resources.spec.js
```diff
@@ -227,6 +231,16 @@ describe('resources', () => {
     checkAllCardsForDataAttribute(DATA_TEST_LANGUAGES, 'JavaScript');
   });
 
+  it('opens and closes login modal', () => {
+    cy.findByTestId(BUTTON).click();
+    cy.findByTestId(MODAL_CONTENT).should('exist').and('be.visible');
+    cy.findAllByTestId(INPUT).should('have.length', 3);
```
kylemh, line 237:
```suggestion
```

I think you're doing enough relevant coverage with just L238. L238 is also less likely to break in the future if we tweak the login modal's content.

components/Modal/Modal.js
```diff
@@ -38,22 +38,23 @@ function Modal({
...
-    >
-      <CloseButton onClick={onRequestClose} />
-      <div
-        className={childrenClassName || ModalStyles.scrollableContainer}
-        data-testid={MODAL_CONTENT}
-      >
-        {children}
-      </div>
-    </ReactModal>
+    <Dialog.Root defaultOpen={false} open={isOpen}>
```
kylemh, line 41: [Notice in the Radix Dialog documentation](https://www.radix-ui.com/docs/primitives/components/dialog), the Dialog button has a bunch of aria properties. I don't actually know if these are needed, since the [WAI modal example](https://www.w3.org/TR/wai-aria-practices-1.2/examples/dialog-modal/dialog.html) has a button _without_ aria attributes. So, I've asked the Radix team. Let me get an answer from them before I give 👍🏼 

```html
<button type="button" aria-haspopup="dialog" aria-expanded="false" aria-controls="radix-30" data-state="closed" class="c-hYXuTU c-hYXuTU-inZbjw-variant-violet">Edit profile</button>
```

cypress/integration/modal.spec.js
florahon: Oh when I looked there already was a test for the close button written right before this overlay one! And that sounds good, I'll switch those out in the other page tests.

components/Modal/Modal.js
kylemh: They strongly encourage using the trigger, but it isn't required to be accessible.

kylemh (approved):

kylemh:
Now just need to make sure tests pass and we'll merge this!

kylemh:
It looks like the close button still isn't visible.

You can check for yourself by going to the resources page and clicking the login button.

Screen Shot 2022-04-19 at 11 53 13 PM

kylemh (changes requested):
It looks like the close button still isn't visible.

You can check for yourself by going to the resources page and clicking the login button.

Screen Shot 2022-04-19 at 11 53 13 PM

florahon:
> It looks like the close button still isn't visible.
> 
> You can check for yourself by going to the resources page and clicking the login button.
> 
> Screen Shot 2022-04-19 at 11 53 13 PM

Hi, sorry about that! I had the close button in the Modal content instead of the overlay. I just pushed the fix.

components/Modal/Modal.js
```diff
@@ -44,11 +44,12 @@ function Modal({
           className={ModalStyles.overlay}
           onClick={shouldCloseOnOverlayClick ? onRequestClose : undefined}
           data-testid={MODAL_OVERLAY}
-        />
+        >
+          <CloseButton theme="secondary" onClick={onRequestClose} />
```
kylemh, line 48: Can we use `https://www.radix-ui.com/docs/primitives/components/dialog#close` with `asChild` prop here?

florahon: I was looking at that earlier - I wasn't sure if we should use it since there's already a CloseButton shared component, but I _think_ it still works if we just add the Dialog Close as a wrapper around it?

florahon: Update - in the example on the Dialog spec, it does look like they're just wrapping existing buttons with the Dialog Close tag, so I can add that quickly.

kylemh: Yep! I wouldn't do it around the definition of CloseButton because the CloseButton could be used in other non-modal contexts. So, just right here is good 😁

components/Modal/Modal.js
```diff
@@ -45,7 +45,9 @@ function Modal({
           onClick={shouldCloseOnOverlayClick ? onRequestClose : undefined}
           data-testid={MODAL_OVERLAY}
         >
-          <CloseButton theme="secondary" onClick={onRequestClose} />
+          <Dialog.Close>
```
kylemh, line 48:
```suggestion
          <Dialog.Close asChild>
```

florahon: I'm so sorry I missed that!!

kylemh:
think the snapshot test is failing

kylemh:
If there is an integration test failing with the modal then forget about Dialog.Trigger and let's get this merged. I need to think more about an ideal API for using Radix Dialog

florahon:
It’s actually the same “join” test failing that I was running into before… I’ve just been seeing that test fail on and off for some reason, but I’m pretty sure none of the new code is doing anything with that side?

components/Modal/__tests__/__snapshots__/Modal.test.js.snap
```diff
@@ -2,66 +2,71 @@
...
+    data-testid="MODAL_OVERLAY"
+    style="pointer-events: auto;"
   >
-    <div
-      class="ReactModal__Overlay"
-      style="position: fixed; top: 0px; left: 0px; right: 0px; bottom: 0px; background-color: rgba(255, 255, 255, 0.75);"
+    <button
+      class="CloseButton"
+      data-testid="CLOSE_BUTTON"
+      type="button"
```
kylemh, line 29: one thing that interests me is that Dialog.Close didn't actually affect the button, so somehow it's not working.

kylemh (approved):
Yay! Let's ship

components/Modal/__tests__/__snapshots__/Modal.test.js.snap
florahon: hmmm what do you mean by that? I'm not sure what adding the wrapper does in terms of affecting any functionality.

kylemh: By doing

```tsx
<Dialog.Close asChild><Button {...props} /* close button */ /></Dialog.Close>
```

What happens is that props are spread onto <Button>, including:
- `aria-haspopup="dialog"`
- `aria-expanded="false"`
- `aria-controls="radix-30"`
- `data-state="closed"`

It's probably because we filter the props in the Button component, but I can fix that later 😁

florahon: Ah that makes sense! Ty!

kylemh:
Wow! Thanks so much for your contribution. It means the world to us here at Operation Code.

As a token of our gratitude, please fill out [this form](https://goo.gl/forms/r02wt0pBNhkxYciI3) to receive some stickers and a handwritten "thank you" note. ❤️

kylemh:
Thanks so much for your hard work! It all should be like in <2m on operationcode.org

florahon:
Hooray!! Thank you for all the help :)

PR #1732, opened 2023-09-17
Add `next-sitemap` dep to correct `<lastmod>` tag values for sitemap entries

### LINKED ISSUE #1719: Sitemap and sitemap generation is broken
# Bug Report
Screenshot 2023-09-05 at 10 36 28 PM

There should be no errors on operationcode.org/sitemap.xml according to Google's Sitemap validator.

AndrewLester:
# Description of changes
This PR adds the `next-sitemap` dependency to correct `<lastmod>` tag values in the sitemap entries. This also means the `sitemap.xml` file is auto generated after build (with the current setup), and the script to generate it is no longer necessary (if I'm not mistaken). Excluded routes and custom priorities are defined in the `next-sitemap.config.js` file.

Lastly, I added the sitemap to the `robots.txt` to inform crawlers, and I added several older links to the sitemap ignore list.

Also, I hardcoded the `siteUrl` (https://www.operationcode.org) in the config, but let me know if there's an environment variable set for this or if I should add that (to a `.env` or something).

# Issue Resolved

Fixes #1719 

## Screenshots/GIFs
image


kylemh (approved):
Amazing work! Thank you.

Issue #1725, opened 2023-09-10
Replace all styling with Tailwind


kylemh:
Blocked by #1724

kylemh:
@subhajit20 You could gradually work on this one! One component and/or one page per PR. It's nice because you won't feel like there's a mountain of work. Just whatever you feel comfortable contributing.

subhajit20:
@kylemh Sure, Can you assign this to me?

kylemh:
Note that you may need to define [Tailwind theme values](https://tailwindcss.com/docs/theme) to leverage the existing CSS values more ergonomically:

https://github.com/OperationCode/front-end/blob/main/common/styles/variables.css
https://github.com/OperationCode/front-end/blob/main/common/styles/media-queries.css
https://github.com/OperationCode/front-end/blob/main/common/styles/themeMap.js

subhajit20:
@kylemh  when I push changes this error is comming 
`Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot`

Will I change ReactDom.render to create room for all the test files?

kylemh:
This is a Next.js application. ReactDOM.render is not used in this repository. Perhaps try re-installing dependencies? Hopefully you've gone through the README instructions!

subhajit20:
> This is a Next.js application. ReactDOM.render is not used in this repository. Perhaps try re-installing dependencies? Hopefully you've gone through the README instructions!

Sorry I have fixed that

akash0708:
Hello @kylemh is the issue still open? I can help in converting the CSS of the footer to Tailwind. Since this is a big issue I hope it wouldn't be a problem with @subhajit20 if I join.

kylemh:
I don't think it'd be a problem! Just steer clear of the work being done in #1746 obviously

akash0708:
Yes, I will make sure I do that. Should I start working on this? Can you assign me?

kylemh:
I can only assign one person to an issue on GitHub. Feel free to work on it regardless. Maybe post here what components you are working on for visibility.

kylemh:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request hopefully this helps!

subhajit20:
[...]

PR #1739, opened 2023-09-19
Converted all the css into Tailwind css of Badge component

subhajit20:
# Description of changes

Converted all the css into Tailwind css of Badge component
Issue - #1725 

# Issue Resolved

Fixes #NA


components/Badge/Badge.module.css
```diff
@@ -1,19 +1,6 @@
...
-  align-items: center;
-  margin: 1rem;
-}
-
 .Badge > img,
 .Badge > svg {
   fill: currentcolor;
   height: 100px;
   margin: 1rem 0;
 }
```
kylemh, line 6: https://csstailwind.com/nested-child-selector-in-tailwindcss-style-nested-elements/

kylemh (commented):
almost done here

components/Badge/Badge.js
```diff
@@ -16,15 +15,20 @@ Badge.defaultProps = {
 
 function Badge({ className, icon, isImageFirst, label }) {
   return (
-    <figure className={classNames(styles.Badge, className)}>
+    <figure
+      className={classNames(
+        'flex flex-col justify-items-center place-content-start m-2 [&>img>svg]:fill-current [&>img>svg]:my-4 [&>img>svg]:mx-0 [&>img>svg]:h-24',
```
kylemh, line 20:
```suggestion
        'flex flex-col justify-items-center place-content-start m-2 [&>img>svg]:fill-current [&>img>svg]:my-4 [&>img]:mx-0 [&>img]:h-24' [&>svg]:mx-0 [&>svg]:h-24',
```

kylemh:
After you finish this PR, perhaps you could integrate Tailwind into our Storybook instance? I forgot about the need to do that!

subhajit20:
> After you finish this PR, perhaps you could integrate Tailwind into our Storybook instance? I forgot about the need to do that!

Okay cool, in the same way I have to do that in stodybook components right?

kylemh:
There shouldn't be any CSS modules to convert in Storybook, it's just adjusting the Storybook config so Tailwind is part of its build process. Right now, storybook has no idea what `m-1` does, for example.

subhajit20:
> There shouldn't be any CSS modules to convert in Storybook, it's just adjusting the Storybook config so Tailwind is part of its build process. Right now, storybook has no idea what `m-1` does, for example.

Okay got it.
Would you merge this pr?

kylemh:
I will as soon as I can! I need to resolve an issue that's visible on this PR's preview deployments on the `/services` route with the Badge components in-use. If I merge without resolving the issue, the problem goes live to prod!

There's nothing you need to do in this PR more though!

subhajit20:
> I will as soon as I can! I need to resolve an issue that's visible on this PR's preview deployments on the `/services` route with the Badge components in-use. If I merge without resolving the issue, the problem goes live to prod!
> 
> There's nothing you need to do in this PR more though!

Alright!

kylemh:
So, @subhajit20 I noticed a problem that you'll need to be aware of...

We've defined the element with the ID `__next` as a point for tailwind to raise its specificity from ([source](https://github.com/OperationCode/front-end/blob/main/tailwind.config.js#L25)) ([tailwind docs on the topic](https://tailwindcss.com/docs/configuration#selector-strategy)).

The downside to this approach is that we need to fixes [like this](https://github.com/OperationCode/front-end/pull/1739/commits/d371d872a8cd636c59dc7e24051d6277480ea9c2#diff-1059fc20b96c24c9675739154b47111f651ccf26b6d4cb2531c74a5ed54166e7) to all the places USING a component we do a Tailwind update to. It's lame, but at least it's formulaic and simple.

We could _not_ do `important: '#__next'` in the Tailwind config; however, if we do that we have the same problem in the other direction... and the solution seems much worse: using the [important selector](https://csstailwind.com/how-to-use-important-modifier-in-tailwindcss-complete-guide/) for all Tailwind classes we want to "win".

The benefit of changing the CSS modules files is that it's temporary, since our end goal has us deleting CSS modules files and the "winner" going forward will just be whatever classname was applied last (ideal for us).

Anyways... to look out for this, you'll want to leverage the "find all references" action in VS Code:

Screenshot 2023-09-20 at 4 55 17 PM

For every component that you work on, you'll want to look for usages, find the page they're rendered on, look at the page and compare the same UI to prod, and either update Tailwind all the way through to the implementation level OR do the fix I recommended above.

PR #1761, opened 2023-11-06
Replaced all the css into tailwind css on Card component

subhajit20:
# Description of changes

Replaced all the css into tailwind css on Card component 
Issue - #1725 

# Issue Resolved

Fixes #NA


kylemh:
Looks like there's still more work to do. Looking at the Vercel build error... it looks like the Modal component imports the Cart CSS.

PR #1797, opened 2024-02-28
Configure Jest with Typescript and convert Button to Typescript.

recondesigns:
# Description of changes

- Configured `.bablerc.js` and `jest.config.js` to test with Typescript files
- Converted Button component to `.tsx`
- Converted Button stories file to `Button.stories.tsx` and updated file to new CSF using Typescript
- Converted `Button.test.js` to `.tsx`and updated snapshot

# Issue Resolved

Fixes #NA

#### Updated file names
Screenshot 2024-02-28 at 5 38 10 PM

#### All tests passing
tests


jest.config.js
```diff
@@ -160,7 +155,7 @@ module.exports = {
   // testLocationInResults: false,
 
   // The glob patterns Jest uses to detect test files
-  testMatch: ['<rootDir>/**/*.test.js'],
+  testMatch: ['<rootDir>/**/*.test.js', '<rootDir>/**/*.test.tsx'],
```
kylemh, line 158: don't forget .ts too!

components/Buttons/Button/Button.tsx
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Sets the button color theme.
+   */
+  theme?: 'primary' | 'secondary';
+  /**
+   * Applies a button type to the base element.
+   */
+  type?: 'button' | 'reset' | 'submit';
+  datum?: any;
+} & React.ButtonHTMLAttributes<HTMLButtonElement>;
```
kylemh, line 77:
```suggestion
type ButtonProps = ComponentProps<'button'> & {
  /**
   * Helps track in-page `event` interactions.
   */
  analyticsObject?: GoogleAnalyticsEventPropType;

   * Forces the component's width as wide as its parent container's width.
   */
  fullWidth?: boolean;

  /**
   * Sets the button color theme.
   */
  theme?: 'primary' | 'secondary';
};
```

components/Buttons/Button/Button.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  return (
+    <button
+      className={classNames(styles.Button, className, styles[theme], {
+        [styles.disabled]: disabled,
+        [styles.fullWidth]: fullWidth,
+      })}
+      data-testid={BUTTON}
+      disabled={disabled}
+      onClick={clickHandler}
```
kylemh, line 116:
```suggestion
  const clickHandler = () => {
   
  };

  return (
    <button
      className={classNames(styles.Button, className, styles[theme], {
        [styles.disabled]: disabled,
        [styles.fullWidth]: fullWidth,
      })}
      data-testid={BUTTON}
      disabled={disabled}
      onClick={(e) => {
        onClick?.(e);
        gtag.event(eventConfig);
      }}
```

prefer defining inline to avoid needing to add explicit types. here we have type safety through implied typing!

kylemh (commented):
just a few changes! great work here!

components/Buttons/Button/Button.tsx
recondesigns: > 

I had to leave onClick in there. Was throwing an error (maybe because of the difference between `onClick` and `onclick`?

components/Buttons/Button/Button.tsx
recondesigns: I still had to apply the `onClick` prop to the type.

jest.config.js
recondesigns: I went ahead and also added .jsx.

components/Buttons/Button/Button.tsx
kylemh: hmmm that doesn't sound right.

components/Buttons/Button/Button.tsx
kylemh: then the typescript setup might not be working

kylemh (approved):

PR #1798, opened 2024-02-29
replaced jest with vitest

### LINKED ISSUE #1727: Replace Jest with Vitest
I want the unit tests to run faster

billy-le:
# Description of changes

Replacing Vitest with Jest for a faster test runner and compatibility with Vite whenever it's implemented

# Issue Resolved

Fixes #1727


kylemh:
You'll have a merge conflict now presumably. I merged the other PR first so I could see that this PR worked with TS first!

kylemh (approved):
Do we need @babel/preset-typescript still? it was added to the babel config. uncertain if that change is still required.

billy-le:
i really don't know. if we start converting cypress js to ts, do you think it will be needed?

kylemh:
Nah. Plan there would be to use Playwright instead of Cypress going forward. Regarding the babel config, I guess it doesn't hurt to have so I'll leave it alone.

PR #1801, opened 2024-03-01
Convert Badge and Branding to Typescript

recondesigns:
# Description of changes

*Converting components included all test and storybook files for each.*

- Converted `Badge` component to Typescript, removed prop-types, and added a props type
- Converted `Branding` components to Typescript, removed prop-types, and added a props type
    - `ColorSection`
    - `FontSection`
    - `LogoSection`
    - `Swatch`

# Issue Resolved

Fixes #NA

#### All new .tsx files
Screenshot 2024-02-29 at 9 58 48 PM


components/Badge/Badge.tsx
```diff
@@ -1,19 +1,25 @@
...
+   */
+  label: string | React.ReactElement;
+  /**
+   * Applies classnames to the base `figure` element for styling.
+   */
+  className?: string;
+  /**
+   * Sets whether the label is rendered above, or below, the badge..
+   */
+  isImageFirst?: boolean;
```
kylemh, line 19:
```suggestion
  /**
   * Sets whether the label is rendered above, or below, the badge...
   * @default - true
   */
  isImageFirst?: boolean;
```

components/Branding/ColorSection/ColorSection.tsx
```diff
@@ -3,7 +3,9 @@ import Swatch from 'components/Branding/Swatch/Swatch';
 import Content from 'components/Content/Content';
 
 function ColorSection() {
+  // @ts-expect-error
```
kylemh, line 6: it's fine to have error expects in test code, but not in UI code... what's the error here?

components/Branding/FontSection/FontSection.tsx
```diff
@@ -1,6 +1,10 @@
 import Content from 'components/Content/Content';
 import { fontsObject } from 'common/styles/styleExports';
 
+// interface FontsObject {
+//   [key: string]: string;
+// }
```
kylemh, line 6: 🚧

kylemh (commented):
looks like we've got a few spots where we're ignoring type errors in app code. Let's resolve those and we should be good to go!

components/Branding/ColorSection/ColorSection.tsx
recondesigns: So, the ts error is: "Property 'primary' does not exist on type '{}'.ts(2339)", and the error flag is on `brandColorsObject.primary`. I think I found a way to solve the error.

Screenshot 2024-03-02 at 8 44 40 AM

kylemh (approved):

PR #1806, opened 2024-03-06
Migrated components to Typescript

recondesigns:
# Description of changes

Migrated the following components to Typescript, removed prop-types, and added a props type:

- `Nav`
    -`NavListItem`
    -`NavMobile`
- `OutboudLink`
- `Pagination`
    - `PaginationItem`
- `PartnerLogoLink`
- `Press`
    - `PressLinks`
    - `PressPhotos`
    - `PressVideos`
- `ProgressIndicator`

*Migrating components included all test and storybook files for each.*

# Issue Resolved

Fixes #NA


components/Press/PressLinks/PressLinks.tsx
```diff
@@ -25,13 +25,15 @@ function PressLinks() {
...
               className={styles.tabsContent}
             >
-              {Articles[region].map(link => (
-                <li key={`GroupLink_${link.url}`}>
-                  <OutboundLink href={link.url} analyticsEventLabel="Press Article">
-                    {link.title}
-                  </OutboundLink>
-                </li>
-              ))}
+              {(Articles as Record<string, any>)[region].map(
```
kylemh, line 28: ugh Object.keys is [annoying](https://fettblog.eu/typescript-better-object-keys/)

Will you please create a utility for Sindre's `objectKeys` fn ([see](https://github.com/sindresorhus/ts-extras/blob/main/source/object-keys.ts)):
`export const objectKeys = Object.keys as <Type extends object>(value: Type) => Array<ObjectKeys<Type>>;`

then you can do `objectKeys(Articles)` on L22 and NOT need to explicitly add types to anything in this file.

kylemh (approved):

utils/types.ts
```diff
@@ -0,0 +1,2 @@
+export const objectKeys = <Type extends object>(value: Type) =>
```
kylemh, line 1: oh you could increase coverage by either ignoring this file or adding a unit test for it.

recondesigns: Do you typically ignore a file like that, or is the convention to create a test for it?

kylemh: I mean it's a pretty basic function created solely for a typing DX improvement. I'm not very worried about it.

PR #1809, opened 2024-03-13
Migrate timeline typescript

recondesigns:
# Description of changes

Migrated the following components to Typescript, removed prop-types, and added a props type:

- `Timeline`
    - `TimelineEvent`
- `UpgradeBrowserOverlay`
    - Added `overlayClassName` to `Modal` props type

*Migrating components included all test and storybook files for each.*

# Issue Resolved

Fixes #NA


components/Timeline/TimelineEvent/TimelineEvent.tsx
```diff
@@ -1,12 +1,11 @@
-import { array, node, oneOfType, string } from 'prop-types';
 import styles from './TimelineEvent.module.css';
 
-TimelineEvent.propTypes = {
-  content: oneOfType([string, array, node]).isRequired,
-  title: string.isRequired,
+export type TimelineEventPropsType = {
+  content: string | string[] | React.ReactNode | React.ReactNode[];
```
kylemh, line 4:
```suggestion
  content: React.ReactNode[];
```

I _think_ this actually covers everything, no?

recondesigns: I thought so, but I was just going off the old prop-type. I need to quit being so skittish about going with my gut lol.

kylemh: I like it better if you assume everything is wrong and just try to make things right :)

PR #1810, opened 2024-03-14
WIP - Migrate Forms to Typescript

recondesigns:
# Description of changes

Migrating `Form` and `Forms` components to Typescript.

# Issue Resolved

Fixes #NA


components/Form/Select/__tests__/Select.test.tsx
```diff
@@ -59,7 +58,7 @@ describe('Select', () => {
       const validate = () => ({ [fieldName]: 'Required' });
 
       const { container, findByText } = render(
-        <Formik initialValues={{ [fieldName]: '' }} validate={validate}>
+        <Formik initialValues={{ [fieldName]: '' }} validate={validate} onSubmit={() => {}}>
```
recondesigns, line 61: Having a hard time with this test (and the rest of the tests in the "interactions" suite. 

I am getting the idea that the Alert is not being rendered in the test, so there is nothing for findByText('Required') to find?

components/Form/Checkbox/Checkbox.tsx
```diff
@@ -1,32 +1,39 @@
...
+   * Sets the name and value for the input element.
+   */
+  field: {
+    name: string;
+    value?: string;
+  };
+  form: {
+    touched: Record<string, any>;
+    errors?: Record<string, string>;
+  };
```
kylemh, line 23: now that we're using TypeScript, we can extend from the actual Formik types!

kylemh: Alternatively, we could not expose those props anymore and leverage the `useField` hook

components/Form/Checkbox/Checkbox.tsx
```diff
@@ -44,7 +51,7 @@ function Checkbox({ field: { name, value, ...field }, form: { errors }, id, labe
       </Label>
 
       <ErrorMessage name={name}>
-        {message => {
+        {(message: string) => {
```
kylemh, line 54: hmmm this shouldn't be needed. If you look in Formik's deps, `ErrorMessage` isn't typed?

components/Form/Checkbox/Checkbox.tsx
kylemh: Yeah, so I guess that's kinda my feedback for all of this PR! Choose your path 😂

components/Form/Checkbox/Checkbox.tsx
recondesigns: If I remove the `string` typing, I just get the TS flag. Not sure how else to get around that.

Screenshot 2024-03-15 at 8 46 26 AM

PR #1893, opened 2025-07-29
convert Accordion to tailwindcss

### LINKED ISSUE #1875: Replace Accordion.tsx styles with TailwindCSS

billy-le:
# Description of changes
Replaces css modules with tailwind. Adjust FAQ page to account for changes.

# Issue Resolved

Fixes #1875 

## Screenshots/GIFs

Screenshot 2025-07-29 at 5 53 22 PM

Screenshot 2025-07-29 at 5 54 38 PM


components/Accordion/Accordion.tsx
```diff
@@ -59,42 +55,32 @@ function Accordion({
...
+          aria-expanded={isContentVisible}
+          className="grid place-items-center w-20 h-20 -mr-6"
+          data-testid={ACCORDION_TOGGLE_BUTTON}
+          id={accordionId}
+          onClick={toggleAccordionContent}
+          type="button"
+        >
+          <ScreenReaderOnly>{toggleMessages[isContentVisible ? 'close' : 'open']}</ScreenReaderOnly>
+          <Chevron
+            className={`w-[30px] fill-current transition-transform duration-100 ease-linear ${isContentVisible ? 'rotate-90' : ''} `}
```
kylemh, line 74: If you define this utility:
```
import { cx as clsx, cva as cvaOriginal } from 'class-variance-authority';
import { twMerge } from 'tailwind-merge';

import type { ClassValue } from 'class-variance-authority/types';

/**
 * Utility function to merge Tailwind CSS classes. This function combines the functionality of
 * `clsx` and `tailwind-merge` to allow for conditional, dynamic application of Tailwind CSS
 * classes while ensuring that conflicting classes are resolved correctly.
 * @see https://github.com/dcastil/tailwind-merge
 * @see https://github.com/lukeed/clsx
 *
 * @example
 * // We should use pseudo-classes when possible, but sometimes we simply need JS for conditional classes.
 * const buttonClass = cx(
 *   'bg-blue-500 hover:bg-blue-700',
 *   { 'underline font-extrabold': shouldBeUnderlinedAndBold },
 *   isLoading && 'animate-spin',
 *   isInViewport ? 'visible' : 'invisible',
 * );
 */
export const cx = (...classes: ClassValue[]) => twMerge(clsx(...classes));

/**
 * This utility is identical in API and result as the "real" CVA. We just use twMerge to resolve
 * class conflicts.
 *
 * @see https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Handling_conflicts
 */
export const cva: typeof cvaOriginal = (base, config) => {
  const cvaFn = cvaOriginal(base, config);
  return (...args: Parameters<typeof cvaFn>) => {
    const result = cvaFn(...args);
    return twMerge(result);
  };
};

export type { VariantProps } from 'class-variance-authority';
```

Then we can use it to maintain the nice conditional syntax that `classNames` has (via `clsx`)

This could then be:
```suggestion
            className={cx('w-[30px] fill-current transition-transform duration-100 ease-linear',  { 
              'rotate-90' : isContentVisible
            })}
```

then I'd add this to `.vscode/settings.json`:
```
"tailwindCSS.classAttributes": ["className"],
  "tailwindCSS.classFunctions": ["cx", "cva", "clsx", "classMerge", "twMerge"],
  "tailwindCSS.lint.cssConflict": "ignore",
  "typescript.preferences.autoImportFileExcludePatterns": [
    "class-variance-authority",
    "clsx",
    "tailwind-merge"
  ]
```

this will make the Tailwind CSS extension work great with all of the tools AND prevent auto-import suggestions from the libraries under the hood of `cx` thereby encouraging people to ALWAYS reach for `cx`.

we could also add a lint rule to ban the imports of `cva`, `clsx`, and `tailwind-merge` outside of the `cx` utility source code file, but I don't have code copy-pasta ready-to-go for that cuz this repo is using an old version of ESLint with a different config system.

billy-le: Sounds good. I went ahead and added the dependency and an eslint rule. I can't vouch for vscode settings since I'm not using that IDE.

For some reason, the eslint is not picking up the restricted import rules. I left it there for the time being. In another PR, I'll go ahead and swap out all the `twMerge` to use the new util

kylemh (approved):
you’ll need to disable comment the lint rule in the actual util file that needs them. i’d also disable the rule until you do that other PR you just suggested

PR #1900, opened 2025-08-01
clean up tw config. update snapshots. replace Button css module with tw

### LINKED ISSUE #1878: Replace Button.tsx styles with TailwindCSS

billy-le:
# Description of changes

Updated config to mirror `variables.css` and modify custom tw variables to be more like tw class names.
Updated all references to new changes.
Converted Button over to cva and tw.

# Issue Resolved

Fixes #1878 

## Screenshots/GIFs

Screenshot 2025-08-01 at 1 18 00 PM
Screenshot 2025-08-01 at 1 18 05 PM
Screenshot 2025-08-01 at 1 18 10 PM
Screenshot 2025-08-01 at 1 18 32 PM
Screenshot 2025-08-01 at 1 18 35 PM
Screenshot 2025-08-01 at 1 18 46 PM
Screenshot 2025-08-01 at 1 19 30 PM
Screenshot 2025-08-01 at 1 19 33 PM


components/Buttons/Button/Button.tsx
```diff
@@ -36,20 +36,53 @@ interface GoogleAnalyticsEventPropType {
...
-   */
-  fullWidth?: boolean;
-  /**
-   * Sets the button color theme.
-   */
-  theme?: 'primary' | 'secondary';
-} & ButtonHTMLAttributes<HTMLButtonElement>;
+}
+
+export const buttonCva = cva(
```
kylemh, line 48: 👏🏼 you're a killer

common/utils/cva.ts
```diff
@@ -36,4 +37,10 @@ export const cva: typeof cvaOriginal = (base, config) => {
...
 };
 
-export type { VariantProps } from 'class-variance-authority';
+type ExcludeNull<T> = { [P in keyof T]: Exclude<T[P], null> };
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export type VariantProps<Comp extends (...args: any) => any> = ExcludeNull<
+  VariantPropsOriginal<Comp>
+>;
+/* eslint-enable @typescript-eslint/no-explicit-any */
```
kylemh, line 46: oh! i see what you're doin. you just need: https://cva.style/docs/getting-started/typescript#extracting-variant-types:~:text=ads%20via%20Carbon-,Extracting%20Variant%20Types,-cva%20offers%20the

```
export const button = cva(/* … */);
export type ButtonProps = VariantProps<typeof button>;
```

right? or does our utility abstraction fn ruin it somehow?

---

if that doesn't work, would `Required<VariantProps<:COMP_IN_QUESTION:>>` not satisfy your needs?

components/Buttons/Button/Button.tsx
```diff
@@ -36,20 +36,53 @@ interface GoogleAnalyticsEventPropType {
...
+  'inline-block font-din-condensed rounded leading-none border-4 border-solid cursor-pointer font-bold text-center uppercase py-4 px-3 whitespace-nowrap transition-all duration-200 ease-linear min-w-[175px]',
+  {
+    variants: {
+      theme: {
+        primary: 'bg-primary border-primary outline-primary text-secondary',
+        secondary: 'bg-secondary border-secondary outline-secondary text-primary',
+      },
+      fullWidth: {
```
kylemh, line 56: later refactor, but i dont think we need this prop/variant anymore... people can easily just do `className="w-full"` now. it was more valuable when CSS Modules

components/Buttons/Button/Button.tsx
```diff
@@ -36,20 +36,53 @@ interface GoogleAnalyticsEventPropType {
...
+        false: null,
+        true: 'w-full',
+      },
+      disabled: {
+        false: null,
+        true: 'opacity-60 hover:cursor-not-allowed',
+      },
```
kylemh, line 63: this shouldn't be a variant.just gotta define `disabled:` rules on L49

```
disabled:opacity-60 disabled:hover:cursor-not-allowed
```

components/Buttons/Button/Button.tsx
```diff
@@ -36,20 +36,53 @@ interface GoogleAnalyticsEventPropType {
...
+    },
+    compoundVariants: [
+      {
+        theme: 'primary',
+        disabled: false,
+        class:
+          'focus-visible:bg-transparent focus-visible:text-primary hover:bg-transparent hover:text-primary',
```
kylemh, line 70: It seems that these are constantly applied, so i'd also move to L49
`focus-visible:bg-transparent hover:bg-transparent`

kylemh (approved):

common/utils/cva.ts
billy-le: I did this because the current version of cva adds `null` and `undefined` to the variants. This will be fixed in the cva@beta according to this issue

https://github.com/joe-bell/cva/issues/253

Doing the required will need us to pass in the props and I think we would want to allow the defaults to take over.

components/Buttons/Button/Button.tsx
```diff
@@ -43,43 +43,34 @@ interface ButtonProps
    * Helps track in-page `event` interactions.
    */
   analyticsObject?: GoogleAnalyticsEventPropType;
+  disabled?: boolean;
```
kylemh, line 46: you can delete this and change the `Omit<ButtonHTMLAttributes<HTMLButtonElement>, 'disabled'>` to `ButtonHTMLAttributes<HTMLButtonElement>`

billy-le: Nice find. I didn't catch that.

kylemh (approved):

PR #1912, opened 2025-08-12
Chore/nav tw

### LINKED ISSUE #1887: Replace Nav.tsx styles with TailwindCSS

billy-le:
# Description of changes

replaces nav css modules to tw. slight visual adjustments due to using tw default classes.

# Issue Resolved

Fixes #1887

## Screenshots/GIFs

Mobile
Left: Changes - Right: Original
Screenshot 2025-08-12 at 10 20 50 AM

Mobile Menu
Screenshot 2025-08-12 at 10 21 01 AM

Desktop
Screenshot 2025-08-12 at 10 23 14 AM

Desktop SubMenu
Screenshot 2025-08-12 at 10 23 49 AM

### DIFF
--- a/common/constants/navigation.ts
+++ b/common/constants/navigation.ts
@@ -1,127 +1,118 @@
-interface NavLink {
-  name: string;
-  href: string;
-  sublinks?: { name: string; href: string }[];
-  isExternal?: boolean;
-}
-
-type MobileNavLink = Pick<NavLink, 'name' | 'href'>;
-
-const donate: NavLink = {
+const donate = {
   href: '/donate',
   name: 'Donate',
 };
 
-const services: NavLink = {
+const services = {
   name: 'Services',
   href: '/services',
 };
 
-const about: NavLink = {
+const about = {
   name: 'About Us',
   href: '/about',
 };
 
-const history: NavLink = {
+const history = {
   name: 'History',
   href: '/history',
 };
 
-const team: NavLink = {
+const team = {
   name: 'Our Team',
   href: '/team',
 };
 
-const contact: NavLink = {
+const contact = {
   name: 'Contact Us',
   href: '/contact',
 };
 
-const faq: NavLink = {
+const faq = {
   name: 'FAQ',
   href: '/faq',
 };
 
-const podcast: NavLink = {
+const podcast = {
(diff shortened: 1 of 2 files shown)


common/constants/navigation.ts
```diff
@@ -112,7 +112,14 @@ export const desktopNavItems = [aboutUsGroup, servicesGroup, getInvolvedGroup];
...
   [about, getInvolved, ...servicesGroup.sublinks, ...getInvolvedGroup.sublinks].map(
-    ({ sublinks = [], ...item }) => [item, sublinks],
+    ({
+      sublinks = [],
+      ...item
+    }: {
+      name: string;
+      href: string;
+      sublinks?: { name: string; href: string }[];
+    }) => [item, sublinks],
```
kylemh, line 122: instead of this (cuz it's technically casting), i think we should make a `NavLink` type and apply it to every chunk.

example:
```ts
type NavLink = {
  name: string;
  href: string;
  sublinks?: { name: string; href: string }[];
}

const press: NavLink = {
  name: 'Press',
  href: '/press',
};
```

billy-le: I wouldn't mind doing this but Typescript is complaining more and eslint is starting to complain too.

Screenshot 2025-08-12 at 1 04 15 PM

I used `...servicesGroup.sublinks!` and eslint tells me i can't use non-null assertions.

Screenshot 2025-08-12 at 1 10 40 PM

To fix it using the `NavLink` type, I will have to use `servicesGroup.sublinks ?? []` even though it's not an undefined property.

On top of that, I still have to use the type casting in the map iterator.

Let me know if you want me to implement the NavLink type this way.

kylemh: well that's cuz `export const mobileNavItems` is:
```
[Pick<NavLink, 'name' | 'href'>, Pick<NavLink, 'sublinks'>][]
```

billy-le: I went with a different approach. Still getting typescript errors regardless so I am still type casting but it's much more clearly defined.

The mobile nav menu doesn't use the sublinks array so I'm not sure why it was there and I used vanilla js to achieve the same results.

PR #1989, opened 2026-03-22
do the thing

kylemh:
# Description of changes

# Issue Resolved

Fixes #NA


kylemh:
nah i dont wanna do this... it's just not worth the flakiness to add platform-specific tests when the platform is so similar... I may add it separately as mobile-specific viewport visual diff tests, but i'd rather do that in Storybook via Chromoium

